PR #6, opened 2016-07-15
Changed the params in the API endpoints to be copies to prevent modif…

mattnguyen1:
Changed the params in the API endpoints to be copies to prevent modifications

### DIFF
--- a/lib/box-client.js
+++ b/lib/box-client.js
@@ -27,7 +27,8 @@
 // ------------------------------------------------------------------------------
 var util = require('util'),
 	errors = require('./util/errors'),
-	httpStatusCodes = require('http-status');
+	httpStatusCodes = require('http-status'),
+	merge = require('lodash').merge;
 
 // API Resource Managers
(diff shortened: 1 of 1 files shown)


mattwiller:
Rather than changing each of the top-level methods, would it make more sense to do the copy in a single place, e.g. at the top of _makeRequest()?

mattnguyen1:
I think my thought process was that people would consume would expect the params to not change at all (which the get, post, put, etc. functions actually do change), but if our goal is to just stop leaking the access token then I think you're right.

bemeric1:
Good callout @mattwiller . Hadn't realized that the appending of headers happens centrally in _makeRequest. While not mutating params would be ideal, what we do mutate it with in get/post/etc is harmless (method&url). I think we care more about is not mutating it to make it insecure for logging (adding auth tokens), which is centrally done in _makeRequest. So, I'd agree we can make the code cleaner by doing this change in _makeRequest.

PR #14, opened 2016-09-12
Folders - Permanently Delete

Trudko:
Added unit test.
Added implementation.
Added documentation.

### DIFF
--- a/lib/managers/folders.js
+++ b/lib/managers/folders.js
@@ -327,7 +327,7 @@ Folders.prototype.getTrashedFolder = function(folderID, qs, callback) {
  * API Endpoint: '/folders/:folderID/trash/'
  * Method: DELETE
  *
- * @param  {type} folderID Box ID of the folder being requested
+ * @param  {string} folderID Box ID of the folder being requested
  * @param  {Function} callback Called with nothing if successful, error otherwise
  * @returns {void}
  */
--- a/lib/managers/folders.js
+++ b/lib/managers/folders.js
@@ -321,7 +321,6 @@ Folders.prototype.getTrashedFolder = function(folderID, qs, callback) {
 	this.client.get(apiPath, params, this.client.defaultResponseHandler(callback));
 };
 
-
 /**
  * Permanently deletes an folder that is in the trash. The item will no longer exist in Box. This action cannot be undone
  *
@@ -330,7 +329,7 @@ Folders.prototype.getTrashedFolder = function(folderID, qs, callback) {
  *
  * @param  {type} folderID Box ID of the folder being requested
  * @param  {Function} callback Called with nothing if successful, error otherwise
- * @return {void}
+ * @returns {void}
  */
 Folders.prototype.deletePermanently = function(folderID, callback) {
 
--- a/lib/managers/folders.js
+++ b/lib/managers/folders.js
@@ -324,7 +324,7 @@ Folders.prototype.getTrashedFolder = function(folderID, qs, callback) {
 /**
  * Permanently deletes an folder that is in the trash. The item will no longer exist in Box. This action cannot be undone
  *
- * API Endpoint: '/folders/:folderID/trash/'
+ * API Endpoint: '/folders/:folderID/trash'
  * Method: DELETE
  *
  * @param  {string} folderID Box ID of the folder being requested
--- a/lib/managers/folders.js
+++ b/lib/managers/folders.js
@@ -323,6 +323,22 @@ Folders.prototype.getTrashedFolder = function(folderID, qs, callback) {
 
 
 /**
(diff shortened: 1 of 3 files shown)


lib/managers/folders.js
```diff
@@ -301,6 +301,21 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
 	this.client.del(apiPath, null, this.client.defaultResponseHandler(callback));
 };
 
+/**
+ * Permanently deletes an folder that is in the trash. The item will no longer exist in Box. This action cannot be undone
+ *
+ * API Endpoint: '/folders/:folderID/trash/'
+ * Method: DELETE
+ *
+ * @param  {type} folderID Box ID of the folder being requested
```
mattwiller: This should have type `{string}`

tests/lib/managers/folders-test.js
```diff
@@ -276,4 +276,20 @@ describe('Folders', function() {
 		});
 	});
 
+	describe('permanentlyDelete()', function() {
```
mattwiller: Should match the method name, `deletePermanently()`

mattwiller:
Looks good, just a couple minor comments.  If you can fix those up, it should be ready to merge!

tests/lib/managers/folders-test.js
```diff
@@ -276,4 +276,20 @@ describe('Folders', function() {
...
+		it('should call BoxClient defaultResponseHandler method with the callback when response is returned', function(done) {
+			sandbox.mock(boxClientFake).expects('defaultResponseHandler').returns(done);
+			sandbox.stub(boxClientFake, 'del').withArgs('/folders/' + FOLDER_ID + '/trash').yieldsAsync();
+			folders.deletePermanently(FOLDER_ID, done);
+		});
+
+		it('should make DELETE call to remove folder permanently', function(done) {
+
+			sandbox.stub(boxClientFake, 'defaultResponseHandler').yields();
+			sandbox.mock(boxClientFake).expects('del').withArgs('/folders/' + FOLDER_ID + '/trash', null);
```
mattwiller: This stub/mock combination isn't quite right; `.yields()` means that `defaultResponseHandler()` will call its function argument immediately when called.  In this test, that argument is `done`, so the test will end immediately on that line.  This is incorrect; `defaultResponseHandler()` should instead return a function and `boxClient.del()` should yield asynchronously to it.  See your above test.

tests/lib/managers/folders-test.js
```diff
@@ -276,4 +276,20 @@ describe('Folders', function() {
 		});
 	});
 
+	describe('deletePermanently()', function() {
+
+		it('should call BoxClient defaultResponseHandler method with the callback when response is returned', function(done) {
+			sandbox.mock(boxClientFake).expects('defaultResponseHandler').returns(done);
```
mattwiller: I would change this to either `.withArgs(done).returns(done)` or simply `.returnsArg(0)` to capture the expected behavior around `defaultResponseHandler()`, namely that is is passed a callback and produces a new callback that should be called once the operation is complete.

lib/managers/folders.js
```diff
@@ -301,6 +301,21 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
 	this.client.del(apiPath, null, this.client.defaultResponseHandler(callback));
 };
 
+/**
+ * Permanently deletes an folder that is in the trash. The item will no longer exist in Box. This action cannot be undone
+ *
+ * API Endpoint: '/folders/:folderID/trash/'
```
mattwiller: Super minor nitpick, but would you mind removing the trailing slash after `/trash` here?

tests/lib/managers/folders-test.js
Trudko: That is what I wanted to ask , many tests are using withArgs('/folders/1234/metadata/global/properties... Why is that? Wouldn't it make sense only if you want to test with different arguments and test different possible responses?

lib/managers/folders.js
Trudko: Sure , don't worry about nitpicks let's do it right. Need to be bit more careful about this.

tests/lib/managers/folders-test.js
mattwiller: We use `.withArgs()` to assert that the mocked function has the correct arguments passed to it.

Trudko: @mattwiller how would the result test look like? All other tests of this kind are in this format, but I think there might be a room for improvement. We might discuss it closer over hangouts.

Trudko: I've used .withArgs(done).returns(done) just to be more explicit.

mattwiller:
Looks good, you'll just need to resolve the merge conflicts.

PR #15, opened 2016-09-12
Folders - Get Trashed Folder

Trudko:
Added unit test.
Added implementation.
Added documentation.


docs/folders.md
```diff
@@ -124,6 +125,17 @@ client.folders.delete('12345', {recursive: true}, callback);
 ```
 
 
+Get a Trashed Folder
+---------------
+
+A folder can be retrieved from the trash with the [`folders.getTrashedFolder(folderID, qs, callback)`](http://opensource.box.com/box-node-sdk/Folders.html#getTrashedFolder) method.
```
mattwiller: Would you mind clarifying that this does not restore an item from the trash?  I think something like "Information about a folder in the trash can be retrieved with the..." might be easier to understand.

lib/managers/folders.js
```diff
@@ -301,6 +303,53 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
 	this.client.del(apiPath, null, this.client.defaultResponseHandler(callback));
 };
 
+/**
+ * Retrieves a folder that has been moved to the trash
+ *
+ * @param  {type} folderID  - The ID of the folder being requested
```
mattwiller: This should have type `{string}`

lib/managers/folders.js
```diff
@@ -301,6 +303,53 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+ * @param {?Object} qs - Additional options can be passed with the request via querystring. Can be left null in most cases.
+ * @param  {type} callback  - Passed the folder information if it was acquired successfully
```
mattwiller: This should have type `{Function}`

lib/managers/folders.js
```diff
@@ -301,6 +303,53 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+ * @returns {void}
+ */
+Folders.prototype.getTrashedFolder = function(folderID, qs, callback) {
+	var params = {
+		qs: qs
+	};
+
+	var apiPath = urlPath(BASE_PATH, folderID, '/trash/');
```
mattwiller: Please omit the trailing slash after `/trash`

lib/managers/folders.js
```diff
@@ -301,6 +303,53 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+	this.client.get(apiPath, params, function(err, response) {
+
+		// Error - No successful request could be made
+		if (err) {
+			callback(err);
+			return;
+		}
+
+		/* eslint-disable callback-return */
+		switch (response.statusCode) {
```
mattwiller: Why not use the `client.defaultResponseHandler()` here?  It should correctly handle the cases you've outlined in this switch — was there something specific you wanted to handle here?

tests/lib/managers/folders-test.js
```diff
@@ -276,4 +276,18 @@ describe('Folders', function() {
...
 	});
 
+	describe('getTrashedFolder()', function() {
+		it('should make GET request to get trashed folder when called', function() {
+			sandbox.stub(boxClientFake, 'defaultResponseHandler');
+			sandbox.mock(boxClientFake).expects('get').withArgs('/folders/' + FOLDER_ID + '/trash', testParamsWithQs);
+			folders.getTrashedFolder(FOLDER_ID, testQS);
+		});
+
+		it('should call BoxClient defaultResponseHandler method with the callback when response is returned', function(done) {
```
mattwiller: This test doesn't match what your code is actually doing

mattwiller:
Posted a few comments — let me know what you think.

lib/managers/folders.js
Trudko: I wanted to handle 400 , folder not in the trash error. With default response handler you would get unexpected API error.

tests/lib/managers/folders-test.js
Trudko: I was hoping you would comment about tests, because the approach with testing is not clear here. Could you please elaborate about the concept behinds tests. I was trying to mimic existing tests in this case it was test for get folder function:
```
	describe('get()', function() {
		it('should make GET request to get folder info when called', function() {
			sandbox.stub(boxClientFake, 'defaultResponseHandler');
			sandbox.mock(boxClientFake).expects('get').withArgs('/folders/1234', testParamsWithQs);
			folders.get(FOLDER_ID, testQS);
		});
```

Trudko:
Thank you for the comments @mattwiller.

lib/managers/folders.js
mattwiller: You don't need to specifically handle this error, the unexpected API response error will use the actual API response to produce a nice message like `Unexpected API Response [404 Not Found] (not_trashed: "Item is not trashed")` by default

tests/lib/managers/folders-test.js
mattwiller: The test you show here is meant to assert that the BoxClient object has its `get()` method called with the correct arguments when `folders.get()` is called.  The tests you have for `getTrashedFolder()` both reference the `defaultResponseHandler()` method, which is not actually used in your method.

lib/managers/folders.js
Trudko: Yes it is handled , but 400 isn't unexpected API response in this case is it? My understanding is that because 400 is explicitly mentioned in documentation under the errors, it should be treated as expected.

tests/lib/managers/folders-test.js
Trudko: I will update this depending on the comment above.

lib/managers/folders.js
mattwiller: In this case, we use "unexpected" to mean any failure case.  This is to differentiate from some endpoints which might have multiple "expected" responses (e.g. a 200 or 302 might both be considered "successful").  However, you probably would not expect to get a 404 here, since that would mean your call was in error.

Trudko: @mattwiller ok so I will treat any error code as unexpected?

lib/managers/folders.js
```diff
@@ -8,7 +8,9 @@
 // ------------------------------------------------------------------------------
 // Requirements
 // ------------------------------------------------------------------------------
-var urlPath = require('../util/url-path');
+var urlPath = require('../util/url-path'),
+	errors = require('../util/errors'),
+	httpStatusCodes = require('http-status');
```
mattwiller: You don't need the `errors` or `httpStatusCodes` dependencies anymore; they can be safely removed.

tests/lib/managers/folders-test.js
mattwiller: This test still needs to be updated.

Trudko: I  added withArgs(done),  is there something else you want to update?

PR #17, opened 2016-09-16
Files - get trashed file

Trudko:

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -149,9 +149,18 @@ client.files.deletePermanently('12345', callback);
 Get a Trashed File
 ---------------
 
-Information about a file in the trash can be retrieved with the [`files.getTrashedFile(fileID, callback)`](http://opensource.box.com/box-node-sdk/Files.html#getTrashedFile) method.
+Information about a file in the trash can be retrieved with the [`files.getTrashedFile(fileID, qs, callback)`](http://opensource.box.com/box-node-sdk/Files.html#getTrashedFile) method.
 ```js
-client.files.getTrashedFile('12345', callback);
+client.files.getTrashedFile('12345', qs, callback);
+```
+
+Requesting information for only the fields you need with the `fields` query
+string parameter can improve performance and reduce the size of the network
+request.
+
+```js
+// Only get information about a few specific fields.
+client.files.getTrashedFile('12345', {fields: 'size,owned_by'}, callback);
 ```
 
 Upload a New Version of a File
--- a/docs/files.md
+++ b/docs/files.md
@@ -149,8 +149,7 @@ client.files.deletePermanently('12345', callback);
 Get a Trashed File
 ---------------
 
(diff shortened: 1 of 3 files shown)


docs/files.md
```diff
@@ -113,6 +114,14 @@ method will move the file to the user's trash.
 ```js
 client.files.delete('12345', callback);
 ```
+Get a Trashed File
+---------------
+
+Information about a file in the trash can be retrieved with the [`files.getTrashedFile(fileID, callback)`](http://opensource.box.com/box-node-sdk/Folders.html#getTrashedFile) method.
```
mattwiller: The URL to the docs should be http://opensource.box.com/box-node-sdk/Files.html#getTrashedFile

lib/managers/files.js
```diff
@@ -560,6 +560,22 @@ Files.prototype.deleteMetadata = function(fileID, scope, template, callback) {
...
+ * Retrieves a file that has been moved to the trash.
+ *
+ * API Endpoint: '/files/:fileID/trash'
+ * Method: GET
+ *
+ * @param {string} fileID - The ID of the file being requested
+ * @param {Function} callback - Called with nothing if successful, error otherwise
+ * @returns {void}
+ */
+Files.prototype.getTrashedFile = function(fileID, callback) {
```
mattwiller: I think this API endpoint accepts the fields parameter, so we probably need to include the ability to pass qs here.

Trudko: [Does it?](https://docs.box.com/reference#get-a-trashed-file)

mattwiller: Yes, it does:

```
12:09 $ curl https://api.box.com/2.0/files/68475254229/trash?fields=id,name -H "Authorization: Bearer ACCESS_TOKEN"
{"type":"file","id":"68475254229","etag":"2","name":"My Document.pdf"}
```

In general, almost all GET requests accept the fields parameter, even if it isn't explicitly mentioned.

Trudko: Added qs parameter and documentation similar to get function documentation.

Trudko:
Conflicts resolved

lib/managers/files.js
```diff
@@ -575,6 +575,27 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
 
+/*
+ * Retrieves a file that has been moved to the trash.
+ *
+ * API Endpoint: '/files/:fileID/trash'
+ * Method: GET
+ *
+ * @param {string} fileID - The ID of the file being requested
+ * @param {?Object} qs - Additional options can be passed with the request via querystring. Can be left null in most cases.
+ * @param {Function} callback - Called with nothing if successful, error otherwise
```
mattwiller: This comment needs to be updated; the callback is called with information about the file when successful

Trudko: Comment updated.

Trudko:
Conflicts are resolved in this branch, can you merge it or is there something I should do?

PR #19, opened 2016-09-16
Folders - Restore folder

Trudko:

### DIFF
--- a/docs/folders.md
+++ b/docs/folders.md
@@ -14,6 +14,7 @@ group, and perform other common folder operations (move, copy, delete, etc.).
 * [Rename a Folder](#rename-a-folder)
 * [Delete a Folder](#delete-a-folder)
 * [Get a Trashed Folder](#get-a-trashed-folder)
+* [Restore a Folder](#restore-a-folder)
 * [Create a Shared Link for a Folder](#create-a-shared-link-for-a-folder)
 * [Get Collaborations for a Folder](#get-collaborations-for-a-folder)
 * [Create Metadata](#create-metadata)
@@ -133,6 +134,16 @@ Information about a folder in the trash can be retrieved with the [`folders.getT
 ```js
 client.folders.getTrashedFolder('12345', {fields: 'name,shared_link,permissions,collections,sync_state'},
 , callback);
+```
+
+
+Restore a Folder
+---------------
+
+A folder can be restored with the [`folders.restoreFolder(folderID, name, parentFolderID qs, callback)`](http://opensource.box.com/box-node-sdk/Folders.html#restoreFolder) method.
+
+```js
+client.folders.restoreFolder('12345', 'New Name', '0', callback);
 ```
 
 
--- a/docs/folders.md
+++ b/docs/folders.md
@@ -14,7 +14,7 @@ group, and perform other common folder operations (move, copy, delete, etc.).
 * [Rename a Folder](#rename-a-folder)
 * [Delete a Folder](#delete-a-folder)
 * [Get a Trashed Folder](#get-a-trashed-folder)
-* [Restore a Folder](#restore-a-folder)
+* [Restore a Folder from Trash](#restore-a-folder-from-trash)
 * [Create a Shared Link for a Folder](#create-a-shared-link-for-a-folder)
 * [Get Collaborations for a Folder](#get-collaborations-for-a-folder)
 * [Create Metadata](#create-metadata)
@@ -137,13 +137,13 @@ client.folders.getTrashedFolder('12345', {fields: 'name,shared_link,permissions,
 ```
 
 
-Restore a Folder
----------------
+Restore a Folder from Trash
+---------------------------
 
-A folder can be restored from the trash with the [`folders.restoreFolder(folderID, name, parentFolderID, callback)`](http://opensource.box.com/box-node-sdk/Folders.html#restoreFolder) method.
+A folder can be restored from the trash with the [`folders.restoreFromTrash(folderID, name, parentFolderID, callback)`](http://opensource.box.com/box-node-sdk/Folders.html#restoreFromTrash) method.
 
 ```js
-client.folders.restoreFolder('12345', 'New Name', '0', callback);
+client.folders.restoreFromTrash('12345', 'New Name', '0', callback);
 ```
 
 
--- a/docs/folders.md
+++ b/docs/folders.md
@@ -140,7 +140,7 @@ client.folders.getTrashedFolder('12345', {fields: 'name,shared_link,permissions,
 Restore a Folder
 ---------------
 
-A folder can be restored from the trash with the [`folders.restoreFolder(folderID, name, parentFolderID qs, callback)`](http://opensource.box.com/box-node-sdk/Folders.html#restoreFolder) method.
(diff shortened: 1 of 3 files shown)


lib/managers/folders.js
```diff
@@ -301,6 +301,35 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+ * Restores an item that has been moved to the trash. Default behavior is to restore the item
+ * to the folder it was in before it was moved to the trash. If that parent folder no longer
+ * exists or if there is now an item with the same name in that parent folder, the new parent
+ * older and/or new name will need to be included in the request.
+ *
+ * API Endpoint: '/folders/:folderID'
+ * Method: POST
+ *
+ * @param {string} folderID - The ID of the folder to restore
+ * @param {string} name - The new name for the folder
```
mattwiller: This parameter is optional!  We denote that in JSDoc by putting square brackets around the param name or by putting a question mark before the type (to indicate nullability).  In this case, since it's possible for name to be left blank but a new parent folder ID to be provided, I think it's best to make the param nullable and optional, e.g. `@param {?string} [name]`

lib/managers/folders.js
```diff
@@ -301,6 +301,35 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+ * @param {string} parentFolderID - The new parent folder for this item
```
mattwiller: See above, but this param is also optional: `@param {string} [parentFolderID]`

lib/managers/folders.js
```diff
@@ -301,6 +301,35 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+ * @returns {void}
+ */
+Folders.prototype.restoreFolder = function(folderID, name, parentFolderID, callback) {
+
+	var params = {
+		body: {
+			name: name,
+			parent: {
+				id: parentFolderID
+			}
```
mattwiller: With the optional params, we'll need to make sure we only add them to the request body if they were passed into this function.

tests/lib/managers/folders-test.js
```diff
@@ -276,4 +276,19 @@ describe('Folders', function() {
 		});
 	});
 
+	describe('restoreFolder()', function() {
+
+		it('should make POST call to restore a folder', function() {
+			sandbox.stub(boxClientFake, 'defaultResponseHandler');
+			sandbox.mock(boxClientFake).expects('post').withArgs('/folders/1234');
```
mattwiller: You should also be asserting that the params object passed in is correct

tests/lib/managers/folders-test.js
```diff
@@ -276,4 +276,19 @@ describe('Folders', function() {
...
+			folders.restoreFolder(FOLDER_ID, 'Folder Restored', 0);
+		});
+
+		it('should call BoxClient defaultResponseHandler method with the callback when response is returned', function(done) {
+			sandbox.mock(boxClientFake).expects('defaultResponseHandler').withArgs(done).returns(done);
+			sandbox.stub(boxClientFake, 'post').yieldsAsync();
+			folders.restoreFolder(FOLDER_ID, 'Folder Restored', 0, done);
+		});
+	});
```
mattwiller: You'll also want to add tests for the optional parameters

Trudko: I did add values as part of the test. Do you want to add tests for different values for the name and / or parentFolderId?

lib/managers/folders.js
```diff
@@ -301,6 +301,45 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+	};
+
+	if (name) {
+		Object.assign(params.body, {
+			name: name
+		});
+	}
+
+	// id can be 0
+	if (parentFolderID != null) {
```
mattwiller: You can simply do `if (parentFolderID)` here, because the ID is a string not a number.

lib/managers/folders.js
```diff
@@ -301,6 +301,45 @@ Folders.prototype.deleteMetadata = function(folderID, scope, template, callback)
...
+		Object.assign(params.body, {
```
mattwiller: Again, I think `Object.assign()` is overkill here.  Why not just do the assignment in one line, e.g. `params.body.parent = {id: parentFolderID};`?

tests/lib/managers/folders-test.js
mattwiller: Yes, we should exercise all the code paths:
- When both optional parameters are passed
- When just name is passed
- When just parentfolderID is passed
- When neither optional parameter is passed

lib/managers/folders.js
Trudko: Wanted to be extra careful if someone does pass 0.

tests/lib/managers/folders-test.js
Trudko: I initialized all of parameters in the each function plus I removed the 'Folder Restored' string and replaced it with name.

lib/managers/folders.js
Trudko: Sure , Object.assign is just personal preference and I saw it being used elsewhere but that was for full fileData object.

docs/folders.md
```diff
@@ -124,6 +125,16 @@ client.folders.delete('12345', {recursive: true}, callback);
 ```
 
 
+Restore a Folder
+---------------
+
+A folder can be restored from the trash with the [`folders.restoreFolder(folderID, name, parentFolderID qs, callback)`](http://opensource.box.com/box-node-sdk/Folders.html#restoreFolder) method.
```
mattwiller: Just need to remove the qs here and this should be good to merge.

lib/managers/folders.js
Trudko: What do you think about this?

docs/folders.md
Trudko: Removed.

mattwiller:
Looks good, just need to resolve merge conflicts.

docs/folders.md
```diff
@@ -14,6 +14,7 @@ group, and perform other common folder operations (move, copy, delete, etc.).
 * [Rename a Folder](#rename-a-folder)
 * [Delete a Folder](#delete-a-folder)
 * [Get a Trashed Folder](#get-a-trashed-folder)
+* [Restore a Folder](#restore-a-folder)
```
mattwiller: We should rename this "Restore a Folder from Trash" to make this more clear

lib/managers/folders.js
```diff
@@ -321,6 +321,40 @@ Folders.prototype.getTrashedFolder = function(folderID, qs, callback) {
...
+ * API Endpoint: '/folders/:folderID'
+ * Method: POST
+ *
+ * @param {string} folderID - The ID of the folder to restore
+ * @param {?string} [name] - The new name for this item
+ * @param {string} [parentFolderID] - The new parent folder for this item
+ * @param {Function} callback - Called with folder information if successful, error otherwise
+ * @returns {void}
+ */
+Folders.prototype.restoreFolder = function(folderID, name, parentFolderID, callback) {
```
mattwiller: I think `restoreFolder` is redundant, since this is already in the Folders manager.  We should rename to `restoreFromTrash` for clarity.

docs/folders.md
Trudko: Renamed.

PR #21, opened 2016-09-19
JWT documentation

Trudko:
Documentation for enterprise and app user authentication.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -126,25 +126,5 @@ var sdk = new BoxSDK({
 	}
 });
 var adminClient = sdk.getEnterpriseAppAuthTokens('YOUR-ENTERPRISE-ID', callback);
-```
-
-### JSON Web Token User Authentication
-
-Once you have enterprise token, you can create App user and then perform
-user-specific actions. See the [Creating App Users and Managing Your Enterprise](https://docs.box.com/docs/app-auth#section-creating-app-users-and-managing-your-enterprise)
-for detailed instructions on how to manage enterprise and perform user based
-actions.
-
-```js
-var BoxSDK = require('box-node-sdk');
-var sdk = new BoxSDK({
-	clientID: 'YOUR-CLIENT-ID',
-	clientSecret: 'YOUR-CLIENT_SECRET',
-	appAuth: {
-		keyID: 'YOUR-KEY-ID',
-		privateKey: 'YOUR-PRIVATE_KEY',
-		passphrase: 'YOUR-PRIVATE-KEY-PASSPHRASE'
-	}
-});
 var client = sdk.getAppUserTokens('APP-USER-ID', callback)
 ```
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -91,6 +91,12 @@ user.  See the [API documentation](https://docs.box.com/docs/getting-started-box
 and [sample app](https://github.com/box/box-node-sdk/blob/master/examples/app-auth)
 for detailed instructions on how to use app auth.
 
+For machine-to-machine use cases, such as applications that use Box services for
+back-end storage, but which provide their own user-inteface, Box extends OAuth 2
+with JSON Web Tokens (JWT). Authentication using JWT enables your application
+to authenticate itself directly to Box without needing to display any Box user
+interface elements.
+
 ```js
 var BoxSDK = require('box-node-sdk');
 var sdk = new BoxSDK({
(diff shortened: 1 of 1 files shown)


docs/authentication.md
```diff
@@ -104,3 +104,47 @@ var sdk = new BoxSDK({
...
+	clientID: 'YOUR-CLIENT-ID',
+	clientSecret: 'YOUR-CLIENT_SECRET',
+	appAuth: {
+		keyID: 'YOUR-KEY-ID',
+		privateKey: 'YOUR-PRIVATE_KEY',
+		passphrase: 'YOUR-PRIVATE-KEY-PASSPHRASE'
+	}
+});
+var adminClient = sdk.getEnterpriseAppAuthTokens('YOUR-ENTERPRISE-ID', callback);
+```
```
mattwiller: This section appears to be mostly a duplicate of the one above; do we need both or can they be combined?

docs/authentication.md
```diff
@@ -104,3 +104,47 @@ var sdk = new BoxSDK({
...
+
+### JSON Web Token User Authentication
```
mattwiller: This shouldn't be a new section, since it's related to App Auth (the section above); at the very least we should add another # to the title to make it a subsection

docs/authentication.md
```diff
@@ -104,3 +104,47 @@ var sdk = new BoxSDK({
...
+
+Once you have enterprise token, you can create App user and then perform
+user-specific actions. See the [Creating App Users and Managing Your Enterprise](https://docs.box.com/docs/app-auth#section-creating-app-users-and-managing-your-enterprise)
```
mattwiller: This documentation page is not really that useful; it describes in detail how to construct a JWT assertion, which the SDK handles for the user.  Why are we including it here?

docs/authentication.md
```diff
@@ -104,3 +104,47 @@ var sdk = new BoxSDK({
...
+var sdk = new BoxSDK({
+	clientID: 'YOUR-CLIENT-ID',
+	clientSecret: 'YOUR-CLIENT_SECRET',
+	appAuth: {
+		keyID: 'YOUR-KEY-ID',
+		privateKey: 'YOUR-PRIVATE_KEY',
+		passphrase: 'YOUR-PRIVATE-KEY-PASSPHRASE'
+	}
+});
+var client = sdk.getAppUserTokens('APP-USER-ID', callback)
```
mattwiller: Since this is mostly the same as the App Auth code example above, maybe they could be combined?

docs/authentication.md
Trudko: I think they might be combined, I wasn't sure if in this case we should just create once section. I can add the line
```js
var client = sdk.getAppUserTokens('APP-USER-ID', callback)
````
to the bottom of top section.

File: docs/authentication.md (hunk 3 above)
Comment (Trudko) [inline docs/authentication.md]: Goal to provide information about relationship between Enterprise user and Application user. Especially that you need to first get enterprise token to receive user id in order to be able to perform user specific action. When I was working with SDK I first used developer token , which obviously works differently than JWT so I had to go back to documentation to look for closer explanation.

File: docs/authentication.md (hunk 1 above)
Comment (mattwiller) [inline docs/authentication.md]: Okay, sounds good!

Comment (Trudko) [inline docs/authentication.md]: OK I've merged it.

**Comment by mattwiller** [comment]
Can we further merge this into the App User section above?  The example code here is nearly exactly the same, so I think it would make sense to have a single section, since App Auth is in fact powered by JWT; they're the same thing.

**Comment by Trudko** [comment]
I've merged App User and machine to machine sections. Added merged commentary and renamed variables to be clear which is which.

File: docs/authentication.md [hunk 5]
```diff
@@ -103,4 +109,7 @@ var sdk = new BoxSDK({
 	}
 });
 var adminClient = sdk.getAppAuthClient('enterprise', 'YOUR-ENTERPRISE-ID');
+// Machine to machine
+var enterpriseApp = sdk.getEnterpriseAppAuthTokens('YOUR-ENTERPRISE-ID', callback);
+var userApp = sdk.getAppUserTokens('APP-USER-ID', callback)
```
mattwiller: Why would the user want to access these methods directly?  It seems more likely to me that we would want to document the process for getting a client (see line 111), since that's how they would make calls to the API.

Trudko: What is purpose of these functions? I didn't find them used anywhere else and they are not marked as @private so I though that they might be used as alternative to AppAuthClient. 

If they are not supposed to be used by user, shouldn't we marked them as private?

PR #24, opened 2016-09-21
Files - Restore Item

Trudko:

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -12,6 +12,7 @@ file's contents, upload new versions, and perform other common file operations
 * [Copy a File](#copy-a-file)
 * [Delete a File](#delete-a-file)
 * [Delete Permanently](#delete-permanently)
+* [Restore Item](#restore-item)
 * [Upload a New Version of a File](#upload-a-new-version-of-a-file)
 * [Download a Previous Version of a File](#download-a-previous-version-of-a-file)
 * [Create a Shared Link](#create-a-shared-link)
@@ -123,6 +124,17 @@ method will remove the file permanently from the user's trash.
 
 ```js
 client.files.deletePermanently('12345', callback);
+```
+
+Restore Item
+-------------
+
+Calling the
+[`files.restoreItem(fileID, name, parentFolderID, callback)`](http://opensource.box.com/box-node-sdk/Files.html#restoreItem)
+will restore an item from the user's trash.
+
+```js
+client.files.restoreItem('12345', 'New Name', 0, callback);
 ```
 
 Upload a New Version of a File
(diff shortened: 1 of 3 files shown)


mattwiller:
This looks good, but you have some merge conflicts you'll need to take care of.

Trudko:
Conflicts resolved.

lib/managers/files.js
```diff
@@ -575,6 +575,43 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
+ * Restores an item that has been moved to the trash. Default behavior is to
+ * restore the item to the folder it was in before it was moved to the trash.
+ * If that parent folder no longer exists or if there is now an item with the
+ * same name in that parent folder, the new parent folder and/or new name will
+ * need to be included in the request.
+ *
+ * API Endpoint: '/files/:fileID'
+ * Method: POST
+ *
+ * @param {string} fileID - The ID of the file to remove metadata from
```
mattwiller: This comment needs to be updated

docs/files.md
```diff
@@ -12,6 +12,7 @@ file's contents, upload new versions, and perform other common file operations
 * [Copy a File](#copy-a-file)
 * [Delete a File](#delete-a-file)
 * [Delete Permanently](#delete-permanently)
+* [Restore Item](#restore-item)
```
mattwiller: Can we rename this to "Restore File From Trash" to be more informative?

lib/managers/files.js
```diff
@@ -575,6 +575,43 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
+ * @param {?string} [name] - The new name for this item
+ * @param {string} [parentFolderID] - The new parent folder for this item
+ * @param {Function} callback - Called with item information if successful, error otherwise
+ * @returns {void}
+ */
+Files.prototype.restoreItem = function(fileID, name, parentFolderID, callback) {
```
mattwiller: See above, but I think we should rename this to restoreFromTrash to be more informative

docs/files.md
Trudko: Sure , updated :)

mattwiller:
I don't see any of your updates, and it looks like you have some merge conflicts.  Would you mind updating and pushing again once those are resolved?

Trudko:
Yes, for the future I will create larger pull requests to prevent most of merge conflicts. Almost all conflicts were caused by merging code on two feature branches within the same file. Larger pull requests should save you and me some work.

mattwiller:
I tend to prefer smaller PRs, so that we don't have huge reviews spanning multiple units of functionality.  How difficult are the merge conflicts to resolve?

Trudko:
I agree with you about not having huge reviews. The conflicts are relatively easy to solve , the problem is that often when I resolve conflicts with master in PR and one of those PR get merged I have another round of conflicts. This might have been specific to files / folders because I put the code mostly at the end of code and that's where conflicts mostly occurred.  

One solution would be to  create PR for  two to three functions instead of just one or that I would resolve conflict in PR which I know will be merged next.

Trudko:
Hi Matt, could you please explain why have you closed  several PR?

PR #27, opened 2016-09-22
Files - get embed link

Trudko:


lib/managers/files.js
```diff
@@ -576,6 +576,29 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
 
 /**
+ * Used to retrieve an expiring URL for creating an embedded preview session.
+ * The URL will expire after 60 seconds and the preview session will expire after 60 minutes.
+ *
+ * API Endpoint: '/files/:fileID?fields=expiring_embed_link'
+ * Method: GET
+ *
+ * @param {string} fileID - The ID of the file to generate embed link for
+ * @param {Function} callback - Called with nothing if successful, error otherwise
```
mattwiller: This comment needs to be updated; the callback will be called with the embed link when successful

lib/managers/files.js
```diff
@@ -576,6 +576,29 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
+Files.prototype.getEmbedLink = function(fileID, callback) {
+
+	var params = {
+		qs: {
+			fields: 'expiring_embed_link'
+		}
+	};
+
+	var apiPath = urlPath(BASE_PATH, fileID);
+	this.client.get(apiPath, params, this.client.defaultResponseHandler(callback));
```
mattwiller: I don't think we should use the default response handler here; rather, perhaps we can just pass the link directly when successful

Trudko: Sure , now only the link it self is passed.

lib/managers/files.js
Trudko: Commend updated.

docs/files.md
```diff
@@ -192,6 +193,17 @@ client.files.getThumbnail('12345', null, function(error, response) {
 });
 ```
 
+Get Embed Link
+--------------
+
+An embed link for a file can be generated by calling the
+[`files.getEmbedLink(fileID,callback)`](http://opensource.box.com/box-node-sdk/Files.html#getEmbedLink)
+method.
```
Trudko: I added simple description.

mattwiller:
It looks like there's some conflicts to resolve here.

PR #28, opened 2016-09-23
Files - lock and unlock

Trudko:
Implementation, unit tests.

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -16,7 +16,8 @@ file's contents, upload new versions, and perform other common file operations
 * [Download a Previous Version of a File](#download-a-previous-version-of-a-file)
 * [Create a Shared Link](#create-a-shared-link)
 * [Get Thumbnail](#get-thumbnail)
-* [Lock and Unlocka a File](#lock-and-unlock-a-file)
+* [Lock a File](#lock-a-file)
+* [Unlock a File](#unlock-a-file)
 * [Create Metadata](#create-metadata)
 * [Get Metadata](#get-metadata)
 * [Update Metadata](#update-metadata)
@@ -193,15 +194,26 @@ client.files.getThumbnail('12345', null, function(error, response) {
 });
 ```
 
-Lock and Unlock a File
-----------------------
+Lock a File
+-----------
 
-A file can be locked and unlocked by call the
-[`files.lockUnlock(fileID, type, expiresAt, isDownloadPrevented, callback)`](http://opensource.box.com/box-node-sdk/Files.html#lockUnlock)
+A file can be locked by calling the
+[`files.lock(fileID, expiresAt, isDownloadPrevented, callback)`](http://opensource.box.com/box-node-sdk/Files.html#lock)
 method.
 
 ```js
-client.files.lockUnlock('12345', 'lock', '2015-12-12T10:55:30-08:00', false, callback);
+client.files.lock('12345', '2015-12-12T10:55:30-08:00', false, callback);
+```
+
+Unlock a File
+-------------
+
+A file can be unlocked by calling the
+[`files.unlock(fileID, callback)`](http://opensource.box.com/box-node-sdk/Files.html#unlock)
+method.
+
+```js
+client.files.unlock('12345', callback);
 ```
 
 Create Metadata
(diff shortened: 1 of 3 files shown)


mattwiller:
I don't really like the idea of having a single control-coupled function for locking and unlocking; could you split this into two functions?  e.g.

```js
client.files.lock(fileID, expirationTime, isDownloadPrevented, callback);
client.files.unlock(fileID, callback);
```

docs/files.md
```diff
@@ -16,6 +16,7 @@ file's contents, upload new versions, and perform other common file operations
 * [Download a Previous Version of a File](#download-a-previous-version-of-a-file)
 * [Create a Shared Link](#create-a-shared-link)
 * [Get Thumbnail](#get-thumbnail)
+* [Lock and Unlocka a File](#lock-and-unlock-a-file)
```
mattwiller: You have a small typo here, "Unlock"

docs/files.md
```diff
@@ -192,6 +193,17 @@ client.files.getThumbnail('12345', null, function(error, response) {
 });
 ```
 
+Lock and Unlock a File
```
mattwiller: This documentation needs to be updated now that the functions are split

lib/managers/files.js
```diff
@@ -575,6 +575,73 @@ Files.prototype.deletePermanently = function(fileID, callback) {
 	this.client.del(apiPath, null, this.client.defaultResponseHandler(callback));
 };
 
+
+ /**
+  * Enum of valid actions for locking and unlocking files
+  *
+  * @enum {LockType}
```
mattwiller: The enum JSDoc documents the type of the values in the enum, so this would be `@enum {string}`

lib/managers/files.js
```diff
@@ -575,6 +575,73 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
+  */
+Files.prototype.lockType = {
```
mattwiller: We typically also expose these sorts of constants on the client itself, and the SDK object; see https://github.com/box/box-node-sdk/blob/master/lib/box-client.js#L356 and https://github.com/box/box-node-sdk/blob/master/lib/box-node-sdk.js#L118

lib/managers/files.js
```diff
@@ -575,6 +575,73 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
+* @param {boolean} [isDownloadPrevented] - Whether or not the file can be downloaded while locked
+* @param {Function} callback - Passed with the locked file information if successful, error otherwise
+* @returns {void}
+*/
+Files.prototype.lock = function(fileID, expiresAt, isDownloadPrevented, callback) {
+	var apiPath = urlPath(BASE_PATH, fileID),
+		params = {
+			body: {
+				lock: {
+					type: 'lock'
```
mattwiller: We should reference the constant here

lib/managers/files.js
```diff
@@ -575,6 +575,73 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
+
+	if (isDownloadPrevented !== null && isDownloadPrevented !== undefined) {
+		params.body.lock.is_download_prevented = isDownloadPrevented;
+	}
+
+	this.client.put(apiPath, params, this.client.defaultResponseHandler(callback));
+};
+
+/**
+* Ulocks a file.
```
mattwiller: Small typo, "Unlocks"

lib/managers/files.js
```diff
@@ -575,6 +575,73 @@ Files.prototype.deletePermanently = function(fileID, callback) {
...
+* @param {string} fileID - The ID of the file to unlock
+* @param {Function} callback - Passed with the unlocked file information if successful, error otherwise
+* @returns {void}
+*/
+Files.prototype.unlock = function(fileID, callback) {
+	var apiPath = urlPath(BASE_PATH, fileID),
+		params = {
+			body: {
+				lock: {
+					type: 'unlock'
```
mattwiller: We should reference the constant here

lib/managers/files.js
Trudko: This will be removed, was used for lockunlock.

Trudko: Actually I was planning to remove this after splitting of the function.

lib/managers/files.js
Trudko: Do we need to expose these in particular? What could be the use case out side of lock and unlock functions?

mattwiller: I think in this case, since the user won't be passing the constants anywhere, these should probably just be local constants in this file, not exposed on the prototype.

PR #30, opened 2016-10-04
Change user login

Trudko:

### DIFF
--- a/docs/users.md
+++ b/docs/users.md
@@ -8,6 +8,7 @@ Users represent an individual's account on Box.
 * [Get Email Aliases](#get-email-aliases)
 * [Add Email Alias](#add-email-alias)
 * [Delete Email Alias](#delete-email-alias)
+* [Change Login](#change-login)
 
 Get the Current User's Information
 ----------------------------------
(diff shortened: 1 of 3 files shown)


mattwiller:
This functionality is already available as `users.update(userID, {login: emailAddress}, callback)`; I'm not sure we need an entirely different method for it, since it's a straightforward update call on the field.

Trudko:
Could you please explain how do you decide when to add separate functions and when not? 

For example handling Get Current User Info and Get User Info by a same function seems clear - you want to get user info.

In this case I would argue that intentions behind changing user login are different enough than intentions behind updating other fields in a similar way how lock/unlock have different intention behind it that just simply updating other file parameter. 

Could you clarify how do you look at this and why for example lock and lock were separate functions from update a file?

mattwiller:
@Trudko To me, the difference is in the complexity of the operation being performed.  For lock and unlock, the user would need to create a nested object, which isn't very intuitive for the operation.  I'd be open to doing the same thing for creating/removing shared links.  But for login, it's a straightforward update of a well-named field.  I'm not sure I understand how updating `login` is different from updating, say, `name`.

Trudko:
@mattwiller thank you for the answer. I agree that if the we decide based on complexity of calling the method , then change user login and change user name are equivalent. So I can remove the implementation and update the comment.

PR #32, opened 2016-10-06
Delete user

Trudko:


Trudko:
Please wait with the review, because yesterday the documentation for Delete User was changed and notify and force parameters were moved to query parameters.

PR #33, opened 2016-10-06
Invite User

Trudko:

### DIFF
--- a/docs/users.md
+++ b/docs/users.md
@@ -5,6 +5,7 @@ Users represent an individual's account on Box.
 
 * [Get the Current User's Information](#get-the-current-users-information)
 * [Update User](#update-user)
+* [Invite User](#invite-user)
 * [Get Email Aliases](#get-email-aliases)
 * [Add Email Alias](#add-email-alias)
 * [Delete Email Alias](#delete-email-alias)
@@ -26,6 +27,15 @@ To update a user call the [`users.update(userID, options, callback)`](http://ope
 
 ```js
 client.users.update('123', {name: 'New Name', job_title: 'New Title', phone: '555-1111'}, callback);
+```
+
+
+Invite User
+-----------
+
+To invite an existing user to join an Enterprise call the [`users.invite(enterpriseID, login, callback)`](http://opensource.box.com/box-node-sdk/Users.html#invite) method.
+```js
+client.users.invite('456', 'user@example.com', callback);
 ```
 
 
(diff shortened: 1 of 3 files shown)


docs/users.md
```diff
@@ -5,6 +5,7 @@ Users represent an individual's account on Box.
 
 * [Get the Current User's Information](#get-the-current-users-information)
 * [Update User](#update-user)
+* [Invite User](#invite-user)
```
mattwiller: We should clarify this as "Invite User to Enterprise", to avoid confusion with inviting a user to collaborate on an item.

docs/users.md
```diff
@@ -29,6 +30,15 @@ client.users.update('123', {name: 'New Name', job_title: 'New Title', phone: '55
 ```
 
 
+Invite User
+-----------
+
+To invite an existing user to join an Enterprise call the [`users.invite(enterpriseID, login, callback)`](http://opensource.box.com/box-node-sdk/Users.html#invite) method.
```
mattwiller: It might be worth adding here that this call requires additional scopes for the calling app, e.g.:

> This method requires the "Manage An Enterprise" scope for the enterprise, which can be enabled within your developer console."

lib/managers/users.js
```diff
@@ -78,6 +79,37 @@ Users.prototype.update = function(id, options, callback) {
...
+ *
+ * API Endpoint: '/invites'
+ * Method: POST
+ *
+ * @param {string} enterpriseID - The enterprise the user will be invited to
+ * @param {string} login - The login of the user that will receive the invitation
+ * @param {Function} callback - Passed the invitation information if it was acquired successfully, error otherwise
+ * @returns {void}
+ */
+Users.prototype.invite = function(enterpriseID, login, callback) {
```
mattwiller: I think we'll want to move this functionality onto a separate Enterprises manager, since it deals exclusively with enterprise functionality that is not available for non-admin users.

docs/users.md
Trudko: Good point I've updated the documentation.

docs/users.md
Trudko: I've added the comment.

lib/managers/users.js
Trudko: Sure does the same apply for Create User for example?[ There is a comment in the code](https://github.com/box/box-node-sdk/blob/master/lib/managers/users.js#L135-L138) which seems to be related to this.

PR #41, opened 2016-10-12
Collections - Create or Update

Trudko:

### DIFF
--- a/docs/collections.md
+++ b/docs/collections.md
@@ -6,6 +6,7 @@ all be in the same folder.
 
 * [Get a User's Collections](#get-a-users-collections)
 * [Get the Items in a Collection](#edit-a-collaboration)
+* [Update the Items in a Collection](#update-the-items-in-a-collection)
 
 Get a User's Collections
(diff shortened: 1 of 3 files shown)


mattwiller:
I don't really think this belongs on the Collections manager, since this method is inherently tied to folders.  In any case, this is already possible via `folders.update()`, so I'm not sure we need another method for it specifically.

Trudko:
Given that [you have mentioned that with lock and unlock ](https://github.com/box/box-node-sdk/pull/30#issuecomment-253370159) user need to create nested object  shouldn't be create or update collection also a  separate function? User has to get ids of collections and create(simple) nested object.

mattwiller:
Hmm, I don't think so; in this case, you're still doing literal assignment.  For instance, the user passes in `['12345', '67890']` and you would be setting body params like:

```js
{
    collections: ['12345', '67890']
}
```

There's no construction happening, the user takes the argument they would pass into a function and just sets it on a top-level property.

PR #43, opened 2016-10-12
Get Enterprise Users

Trudko:


Trudko:
This might be also moved to separate Enterprise Manager as you've proposed with Invite User.

mattwiller:
Yeah, I think this would make more sense as part of an Enterprises manager.

PR #44, opened 2016-10-17
Create a task

Trudko:
Implementation, documentation and unit test

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -8,13 +8,15 @@ Tasks enable file-centric workflows in Box. User can create tasks on files and a
 Create a Task
 -------------
 
-To create a task call the [`tasks.create(fileID, message, dueAt, callback)`](http://opensource.box.com/box-node-sdk/Tasks.html#create) method.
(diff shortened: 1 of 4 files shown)


docs/tasks.md
```diff
@@ -0,0 +1,20 @@
+Tasks
+=====
+
+Tasks enabled file-centric workflows in Box. User can create tasks on files and assign them to collaborators on Box.
```
mattwiller: The wording here should be "Tasks enable"

lib/managers/tasks.js
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Used to create a single task for single user on a single file.
+ *
+ * API Endpoint: '/tasks'
+ * Method: POST
+ *
+ * @param {String} fileID - The ID of the item this task is for
+ * @param {String} message - An optional message to include with the task
+ * @param {String} dueAt - The day at which this task is due
```
mattwiller: These two params are optional, they should be moved into an options object and documented

docs/tasks.md
Trudko: I've fixed that, but this needs to be fixed i[n the official API ](https://docs.box.com/reference#task-object-1)reference when I got this from.

PR #46, opened 2016-10-26
Allow rename on copy

### LINKED ISSUE #40: Support for customizing folder/file name on Copy
I'd like to see support for customizing the names of items during a Copy process. The Box API allows for specifying the new item name as an optional parameter in the request (see below), but this is not currently supported in the SDK. Can an optional parameter be added to the Folder Copy and File Copy functions to support this?

Folder Copy: https://docs.box.com/reference#copy-a-folder
File Copy: https://docs.box.com/reference#copy-a-file

Thanks!

mattwiller:
Fixes #40.  When copying a file or folder, an optional parameter
can be passed to perform a rename when copying.  This commit adds a
new options argument to files.copy() and folders.copy() to support
such a parameter.

/cc: @djordan

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -124,6 +124,14 @@ method.
 client.files.copy('12345', '0', callback);
 ```
 
+An optional `name` parameter can also be passed to rename the file on copy.  This can be
+used to avoid a name conflict when there is already an item with the same name in the
+target folder.
+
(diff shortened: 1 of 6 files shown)


djordan:
Looks good, @mattwiller. 🚀

PR #53, opened 2016-11-01
Add Legal Holds functionality

mattwiller:
/cc: @djordan


Trudko:
Hi Matt, I've noticed that you are passing options objects directly into request body so names of object parameters are same as are in the API.  [Create Task PR](https://github.com/box/box-node-sdk/pull/44/files#diff-234cc401d5d2e1b3e77c6c20432f38a0R63) which you merged does map object parameters to API params. We have inconsistent API.

Reason I provided mapping is so SDK api has naming convention as JS usually does. IIRC other SDKs do provide this mapping.

mattwiller:
@Trudko Yeah, we never really had a standard before.  I spoke to some of the other contributors, and we think it probably makes the most sense to just pass the options object directly.  This way, when new parameters are added to the API users can just pass them directly to the options object without us needing to update the SDK.  I'll go ahead and update all the existing code to conform to this standard, but going forward we should plan on sticking to this one way.

Trudko:
@mattwiller ok I can update my existing code. We should just keep in mind that with the new parameters we still have to update JSDoc.

PS: What do you think about an idea of allowing users to pass callback instead if options object instead of forcing them to pass null? It's bit of extra work on our side, but it might make SDK slightly friendlier to work on.

Trudko:
Three more questions:

1. Why isn't options object documented for update function?
2. Why isn't qs object documented for get function? I know that all of qs have fields param, but still could be nice to add it here too.
3. What about documenting default values?

PR #56, opened 2016-11-03
Update Web Link

Trudko:

### DIFF
--- a/lib/managers/web-links.js
+++ b/lib/managers/web-links.js
@@ -106,13 +106,7 @@ WebLinks.prototype.update = function(weblinkID, options, callback) {
 			}
 		};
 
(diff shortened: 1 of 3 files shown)


lib/managers/web-links.js
```diff
@@ -86,4 +86,35 @@ WebLinks.prototype.get = function(weblinkID, qs, callback) {
...
+			}
+		};
+
+	if (options && options.name) {
+		params.body.name = options.name;
+	}
+
+	if (options && options.description) {
+		params.body.description = options.description;
+	}
```
mattwiller: These should be updated to the standard options passing style, where all options are passed in the body

Trudko:
Code updated.

PR #59, opened 2016-11-08
Create User

Trudko:


Trudko:
I added also default  values to JS Doc. Let me know what do you think.

PR #67, opened 2016-11-10
Get Watermark On Folder

Trudko:
Implementatiom, documentation and unit tests.


Trudko:
I added qs parameter but so far it doesn't seem to be doing much. Is it or will it be supported or should I remove it?

PR #69, opened 2016-11-11
Add Groups functionality

mattwiller:
/cc: @djordan

### DIFF
--- a/lib/managers/groups.js
+++ b/lib/managers/groups.js
@@ -1,0 +1,321 @@
+/**
+ * @fileoverview Manager for the Groups resource
+ * @author mwiller
+ */
+
+'use strict';
+
+// -----------------------------------------------------------------------------
+// Typedefs
+// -----------------------------------------------------------------------------
+
+/**
+ * Describes what types of users can perform certain actions on a group
+ * @typedef {string} GroupAccessLevel
+ */
+
+/**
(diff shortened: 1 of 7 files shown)


lib/managers/groups.js
```diff
@@ -0,0 +1,321 @@
...
+
+/**
+ * Fetch a specific membership record, which shows that a given user is a member
+ * of some group.
+ *
+ * API Endpoint: '/group_memberships/:membershipID'
+ * Method: GET
+ *
+ * @param {string} membershipID - The ID of the membership to fetch
+ * @param {?Object} qs - Optional parameters, can be left null in most cases
```
djordan: I realize the name is different because it's a get, but should we call all of these "options"?

lib/managers/groups.js
```diff
@@ -0,0 +1,321 @@
...
+
+/**
+ * Retreieve a list of memberships for the group, which show which users
+ * belong to the group
+ *
+ * API Endpoint: '/groups/:groupID/memberships'
+ * Method: GET
+ *
+ * @param {string} groupID - The ID of the group to get memberships for
+ * @param {?Object} options - Optional parameters, can be left null in most cases
```
djordan: Hmmmm... here we have it as options, not qs.

PR #70, opened 2016-11-11
Add Files methods

mattwiller:
Implement lock/unlock, restore from trash, get versions, and delete
version.

/cc: @djordan

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -14,12 +14,17 @@ file's contents, upload new versions, and perform other common file operations
 * [Delete a File](#delete-a-file)
 * [Delete Permanently](#delete-permanently)
 * [Get a Trashed File](#get-a-trashed-file)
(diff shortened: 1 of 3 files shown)


docs/files.md
```diff
@@ -263,6 +298,57 @@ For more information, see the [API documentation](https://docs.box.com/reference
 client.files.getEmbedLink('12345', callback);
 ```
 
+Lock a File
+-----------
+
+A file can be locked, which prevents other users from editing the file, by calling the
+[`files.lock(fileID, expiresAt, isDownloadPrevented, callback)`](http://opensource.box.com/box-node-sdk/Files.html#lock)
```
djordan: Looks like you put expiresAt and isDownloadPrevented into an options object. The example below looks OK, but the signature above is wrong.

PR #72, opened 2016-11-11
Delete Task

Trudko:

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -5,6 +5,7 @@ Tasks enable file-centric workflows in Box. User can create tasks on files and a
 
 * [Create a Task](#create-a-task)
 * [Get a Task's Information](#get-a-tasks-information)
+* [Delete a Task Permanently](#delete-a-task-permanently)
 
 Create a Task
 -------------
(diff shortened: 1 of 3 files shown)


Trudko:
Can I create one larger PR for rest of tasks so you don't have review it one by one and I don't have to resolve merge conflicts over and over?

docs/tasks.md
```diff
@@ -5,6 +5,7 @@ Tasks enable file-centric workflows in Box. User can create tasks on files and a
 
 * [Create a Task](#create-a-task)
 * [Get a Task's Information](#get-a-tasks-information)
+* [Delete a Task Permanently](#delete-a-task-permanently)
```
mattwiller: Tasks cannot be temporarily deleted; why do we need to add this qualifier throughout?

Trudko: All other functions which move items to trash  are called delete and those who removed items permanently are called deletePermanently.  I wanted to be consistent so user accidentally doesn't use deleteTask thinking it works same as other delete functions. It might clear for people familiar with sdk but not to new people. If you want to rename it I would at least improve documentation.

PR #77, opened 2016-11-14
Clean up webhooks support

mattwiller:
/cc @djordan

### DIFF
--- a/lib/managers/webhooks.js
+++ b/lib/managers/webhooks.js
@@ -1,50 +1,104 @@
 /**
-* @fileoverview Webhook for the Box Webhooks Resource
-*/
+ * @fileoverview Manager for the Box Webhooks resource
+ */
 'use strict';
 
-// ------------------------------------------------------------------------------
+// -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lib/managers/webhooks.js
```diff
@@ -62,16 +116,16 @@ Webhooks.prototype.create = function(targetID, targetType, notificationURL, trig
...
+ *
+ * API Endpoint: '/webhooks/:webhookID'
+ * Method: GET
+ *
+ * @param {string} webhookID - ID of the webhook to retrieve
+ * @param {?Object} qs - Additional options can be passed with the request via querystring. Can be left null in most cases.
+ * @param {Function} callback - Passed the webhook information if it was acquired successfully
+ * @returns {void}
+ */
 Webhooks.prototype.get = function(webhookID, qs, callback) {
```
djordan: Would be nice to rename "qs" to "options" for consistency.

mattwiller: Yeah, I think I'll just do one big pass and rename everything in another PR

PR #81, opened 2016-11-16
Add token exchange

mattwiller:
Not ready to merge this yet, just wanted to get feedback on the core implementation and where this might fit in the SDK.  Where should we expose this functionality on the client?  Maybe `client.exchangeToken()` — thoughts?

/cc @djordan


mself:
I guess the typical use would be to pass the down-scoped token to another service (typically from server to browser, but Apple wants to pass one from mobile/desktop client to their collaboration server).  If you wanted to create a Node client with the new token could you call `sdk.getBasicClient()` with it?

Overall, looking good!  Before publishing, we may need to wait for the Dev-Exp team to implement "semantic scopes" or some mechanism to allow you to actually get a lower-scoped token from a higher-scoped one.  That's a bit of a gap at the moment.

mattwiller:
@mself Yep, that's exactly why we're sitting on this one for now.  You would be able to do `sdk.getBasicClient()` with the new token or pass it to another service, though — either should be fine.

lib/box-client.js
```diff
@@ -441,13 +433,30 @@ BoxClient.prototype.asSelf = function() {
...
+
+	return Promise.fromCallback(cb => this._session.revokeTokens(cb))
+		.catch(err => errors.unwrapAndThrow(err))
+		.asCallback(callback);
+};
+
+/**
+ * Exchange the client access token for one with lower scope
+ * @param {string|string[]} scopes The scope(s) requested for the new token
+ * @param {string} [resource] The absolute URL of an API resource to scope the new token to
```
djordan: What resources can be specified here? Is there documentation on it somewhere?

tests/lib/box-client-test.js
```diff
@@ -788,16 +862,14 @@ describe('box-client', function() {
 
 		it('should pass unexpected response error to callback when called with unsuccessful response', function(done) {
 
-			var unexpectedResponseError = new Error('dead dove, do not eat'),
-				unexpctedResponse = {
-					statusCode: 403
-				};
-
-			sandbox.stub(errorsFake, 'buildUnexpectedResponseError').returns(unexpectedResponseError);
+			var unexpctedResponse = {
```
djordan: Typo: 'unexpcted'

PR #83, opened 2016-11-16
Restore Folder from Trash

mattwiller:
/cc @djordan


mattwiller:
@djordan Good point; let's follow up tomorrow on whether these should be moved.

PR #88, opened 2016-11-17
As-User support

mattwiller:
/cc @djordan


mattwiller:
@djordan I didn't really know where to document it yet; I'll consult the other SDK owners and see what they think.

PR #95, opened 2016-11-25
Expired Auth Code

gforster1993:
Hello,

Great work on this SDK. We're using the app-auth client and I've been having an issue where I'm getting the error ` Error: Expired Auth: Auth code or refresh token has expired`. I've had a look around and the problem was originating for the token request https://api.box.com/oauth2/token. Not entirely sure why but it appears to be due to the iat (issued at time) parameter in the claims of the JWT assertion.

After adding the noTimestamp flag the assertion no longer contains the iat parameter and I don't have any issues. Maybe you could shed some light onto why this is the case but as it's an optional parameter anyhow I don't see the need for it.

Cheers,
Gary

### DIFF
--- a/lib/token-manager.js
+++ b/lib/token-manager.js
@@ -314,6 +314,7 @@ TokenManager.prototype = {
 			subject: id,
 			issuer: this.config.clientID,
 			jwtid: uuid.v4(),
+			noTimestamp: true,
 			expiresIn: this.config.appAuth.expirationTime,
 			headers: {
 				kid: this.config.appAuth.keyID


mjanathk:
Hi,
 I am running into the same problem. Is there  work around. I tried adding noTimeStamp=true in token-manager.js but it does not work... same error

`[grunt-develop] > { Error: Expired Auth: Auth code or refresh token has expired.
    at Object.buildResponseError (/home/awsgui/service/node_modules/box-node-sdk/lib/util/errors.js:58:23)
    at Object.buildExpiredAuthError (/home/awsgui/service/node_modules/box-node-sdk/lib/util/errors.js:71:28)
    at /home/awsgui/service/node_modules/box-node-sdk/lib/token-manager.js:213:21
    at /home/awsgui/service/node_modules/box-node-sdk/lib/api-request.js:286:3
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9)
  statusCode: 400,
`

mattwiller:
@gforster1993 @mjanathk Could you give me some more details about the errors you're seeing?  Specifically, if you could give me the contents of the `error.response.body` from the error object, it would be really helpful in figuring out how to proceed on this issue.

mjanathk:
@mattwiller,
                    I saw this problem only on my Dev server. The system time on the Dev server that I was working on was off by 1 minute. I did the ntp update and tried again. It worked fine.

mattwiller:
@mjanathk That's what I suspected, that your client's time is ahead of Box servers' and the API is refusing to honor the JWT before the `iat` time.  I think for security reasons we might like to leave including the timestamp as the default, but we could add a configuration option to remove it.  I'll check to see which route would be best.

mattwiller:
@gforster1993 I checked with some other Box SDK maintainers and we would like to include passing the `iat` claim as an optional parameter in the SDK Config object for app auth.  I was thinking something like `verifyTimestamp` which would default to false but could be set to true if the user wanted to use `iat`.  If you'd like to update this PR, let me know, otherwise I can just make the change!

mattwiller:
@gforster1993 I didn't hear back from you, so I went ahead and implemented the fix with an optional parameter to use `iat` — see #99 

I'm going to close this PR, since the new one supersedes it.

PR #99, opened 2016-12-09
Make JWT iat parameter optional

### LINKED ISSUE #96: Intermittent "auth code or refresh token expired" error with App Auth client
screen shot 2016-12-07 at 12 45 03 pm

Hi there,

My application ocassionally gets this error when calling methods on the box client object that is returned when calling `sdk.getAppAuthClient('user', myAppUserId);`. Is this because I need to create a persistent client? How would I create a persistent app auth client?

This error is intermittent and resolves itself in a seemingly arbitrary amount of time. But it would be more ideal to prevent this error - because until the error is resolved my application is unable to use any of the box client methods.

### LINKED ISSUE #97: Auth code expired
I am generated authentication using enterpise like the given syntax
sdk.getAppAuthClient('enterprise', id);

It works some time and some time it throws error like auth is expired and can you help me to resolve this issue i am struck in that issue

mattwiller:
When the iat parameter for JWT token grants is included by default,
many users are encountering errors due to clock skew.  Instead, the
parameter will be optional.

Fixes #96 
Fixes #97

### DIFF
--- a/lib/token-manager.js
+++ b/lib/token-manager.js
@@ -315,6 +315,7 @@ TokenManager.prototype = {
 			issuer: this.config.clientID,
 			jwtid: uuid.v4(),
 			expiresIn: this.config.appAuth.expirationTime,
+			noTimestamp: !this.config.appAuth.verifyTimestamp,
 			headers: {
 				kid: this.config.appAuth.keyID
 			}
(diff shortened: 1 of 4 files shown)


lib/token-manager.js
```diff
@@ -315,6 +315,7 @@ TokenManager.prototype = {
 			issuer: this.config.clientID,
 			jwtid: uuid.v4(),
 			expiresIn: this.config.appAuth.expirationTime,
+			noTimestamp: !this.config.appAuth.verifyTimestamp,
```
djordan: Thanks for making the config option a positive flag.

PR #101, opened 2017-01-18
Added helper function to validate signatures on webhook messages

mself:
Consumers can validate a webhook message by using:
```
const BoxSDK = require('box-node-sdk');
if(BoxSDK.webhooks.validateMessage(body,headers, primarySignatureKey, secondarySignatureKey)) {...}
```

### DIFF
--- a/lib/box-node-sdk.js
+++ b/lib/box-node-sdk.js
@@ -59,7 +59,8 @@ var EventEmitter = require('events').EventEmitter,
 	PersistentAPISession = require('./sessions/persistent-session'),
 	AnonymousAPISession = require('./sessions/anonymous-session'),
 	AppAuthSession = require('./sessions/app-auth-session'),
-	BoxClient = require('./box-client');
+	BoxClient = require('./box-client'),
+	Webhooks = require('./managers/webhooks');
 
 // ------------------------------------------------------------------------------
 // Private
@@ -241,6 +242,13 @@ BoxSDKNode.prototype.getAppUserTokens = function(userID, callback) {
 BoxSDKNode.prototype.revokeTokens = function(token, callback) {
 	this.tokenManager.revokeTokens(token, callback);
 };
+
+/**
+ * Expose the Webhooks module to the SDK as a whole. This allows
+ * the consumer to call BoxSDK.webhooks.validateMessage() by requiring the SDK,
+ * instead of needing to create a client (which is not needed to validate messages).
+ */
+BoxSDKNode.webhooks = Webhooks;
 
 /** @module box-node-sdk/lib/box-node-sdk */
 module.exports = BoxSDKNode;
--- a/lib/box-node-sdk.js
+++ b/lib/box-node-sdk.js
@@ -244,11 +244,11 @@ BoxSDKNode.prototype.revokeTokens = function(token, callback) {
 };
 
 /**
- * Expose the Webhooks module to the SDK as a whole. This allows
- * the consumer to call BoxSDK.webhooks.validateMessage() by requiring the SDK,
+ * Expose Webhooks.validateMessage() to the SDK as a whole. This allows
+ * the consumer to call BoxSDK.validateWebhookMessage() by just requiring the SDK,
  * instead of needing to create a client (which is not needed to validate messages).
  */
-BoxSDKNode.webhooks = Webhooks;
+BoxSDKNode.validateWebhookMessage = Webhooks.validateMessage;
 
 /** @module box-node-sdk/lib/box-node-sdk */
 module.exports = BoxSDKNode;
--- a/lib/managers/webhooks.js
+++ b/lib/managers/webhooks.js
@@ -300,6 +300,8 @@ Webhooks.validateMessage = function(body, headers, primarySignatureKey, secondar
 	return true;
 };
 
+Webhooks.prototype.validateMessage = Webhooks.validateMessage;
+
 /**
  * @module box-node-sdk/lib/managers/webhooks
  * @see {@Link Webhooks}
--- a/lib/managers/webhooks.js
+++ b/lib/managers/webhooks.js
@@ -45,12 +45,10 @@ function computeSignature(body, headers, signatureKey) {
 	}
 
 	if (headers['box-signature-version'] !== '1') {
-		// console.log('Unrecognized signature version: %s', headers['box-signature-version']);
 		return null;
 	}
 
 	if (headers['box-signature-algorithm'] !== 'HmacSHA256') {
-		// console.log('Unrecognized signature algorithm: %s', headers['box-signature-algorithm']);
 		return null;
 	}
 
@@ -59,7 +57,6 @@ function computeSignature(body, headers, signatureKey) {
 	hmac.update(headers['box-delivery-timestamp']);
 
 	const signature = hmac.digest('base64');
-	// console.log('Signature: %s', signature);
 
 	return signature;
 }
@@ -81,18 +78,15 @@ function validateSignature(body, headers, primarySignatureKey, secondarySignatur
 	const primarySignature = computeSignature(body, headers, primarySignatureKey);
 
 	if (primarySignature && primarySignature === headers['box-signature-primary']) {
-		// console.log('Primary signature verified');
 		return true;
 	}
 
 	const secondarySignature = computeSignature(body, headers, secondarySignatureKey);
 
 	if (secondarySignature && secondarySignature === headers['box-signature-secondary']) {
-		// console.log('Secondary signature verified');
 		return true;
 	}
 
-	// console.log('Signature not verified');
 	return false;
 }
 
@@ -109,10 +103,7 @@ function validateDeliveryTimestamp(headers, maxMessageAge) {
 	const currentTime = Date.now();
 	const messageAge = (currentTime - deliveryTime) / 1000;
 
-	// console.log('Message age: %d', messageAge);
-
 	if (messageAge > maxMessageAge) {
-		// console.log('Message too old');
 		return false;
 	}
 
--- a/lib/managers/webhooks.js
+++ b/lib/managers/webhooks.js
@@ -17,6 +17,7 @@
 // -----------------------------------------------------------------------------
 
 var urlPath = require('../util/url-path');
+const crypto = require('crypto');
 
 // -----------------------------------------------------------------------------
 // Private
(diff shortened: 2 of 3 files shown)


lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
 	this.client.del(apiPath, null, this.client.defaultResponseHandler(callback));
 };
 
+const crypto = require('crypto');
+
+/**
+ * Compute the message signature (see https://docs.box.com/reference#signatures)
```
mattwiller: You can use `@see` to expose the docs link better in JSDoc, I think

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+	return signature;
+}
+
+/**
+ * Validate the message signature (see https://docs.box.com/reference#signatures)
+ *
+ * @param {string} body - The request body of the webhook message
+ * @param {Object} headers - The request headers of the webhook message
+ * @param {string} primarySignatureKey - The primary signature to verify the message with
+ * @param {?string} secondarySignatureKey - The secondary signature to verify the message with
```
mattwiller: This param isn't just nullable, it's totally optional!  You can denote that like `@param {string} [secondarySignatureKey]`

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+	if (messageAge > maxMessageAge) {
+		// console.log('Message too old');
+		return false;
+	}
+
+	return true;
+}
+
+// This prevents replay attacks
+const maxMessageAge = 10 * 60; // 10 minutes
```
mattwiller: This should be a constant (UPPER_SNAKE_CASE) at the top of the file

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+/**
+ * Validate a webhook message by verifying the signature and the delivery timestamp
+ *
+ * @param {string} body - The request body of the webhook message
+ * @param {Object} headers - The request headers of the webhook message
+ * @param {string} primarySignatureKey - The primary signature to verify the message with
+ * @param {?string} secondarySignatureKey - The secondary signature to verify the message with
+ * @returns {boolean} - True or false
+ */
+Webhooks.validateMessage = function(body, headers, primarySignatureKey, secondarySignatureKey) {
```
mattwiller: We might want to consider making the timestamp delta configurable; we've run into issues before where clients' system clocks are way out of sync with Box

tests/lib/managers/webhooks-test.js
```diff
@@ -12,6 +12,8 @@ var sinon = require('sinon'),
 
 var BoxClient = require('../../../lib/box-client');
 
+var assert = require('assert');
```
mattwiller: This should go with the rest of the test library require() statements above.  Also, we use Chai for assertions, so this should be `require('chai').assert`

lib/box-node-sdk.js
```diff
@@ -242,5 +243,12 @@ BoxSDKNode.prototype.revokeTokens = function(token, callback) {
 	this.tokenManager.revokeTokens(token, callback);
 };
 
+/**
+ * Expose the Webhooks module to the SDK as a whole. This allows
+ * the consumer to call BoxSDK.webhooks.validateMessage() by requiring the SDK,
+ * instead of needing to create a client (which is not needed to validate messages).
+ */
+BoxSDKNode.webhooks = Webhooks;
```
mattwiller: I'm not quite sure how we want to expose this.  Maybe have a `Webhooks.helpers` on the Webhooks module, and expose it here as `BoxSDKNode.webhooks`?

lib/managers/webhooks.js
```diff
@@ -179,7 +179,7 @@ Webhooks.prototype.update = function(webhookID, options, callback) {
 };
 
 /**
- * Delete a specifed webhook by ID
+ * Delete a specified webhook by ID
```
mself: :-)

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
 	this.client.del(apiPath, null, this.client.defaultResponseHandler(callback));
 };
 
+const crypto = require('crypto');
```
mself: done

lib/managers/webhooks.js
mself: done

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+
+/**
+ * Compute the message signature (see https://docs.box.com/reference#signatures)
+ *
+ * @param {string} body - The request body of the webhook message
+ * @param {Object} headers - The request headers of the webhook message
+ * @param {string} signatureKey - The signature to verify the message with
+ * @returns {string} - The message signature
```
mself: done

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+ * @private
+ */
+function computeSignature(body, headers, signatureKey) {
+	if (!signatureKey) {
+		return undefined;
```
mself: done

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+	}
+
+	if (headers['box-signature-version'] !== '1') {
+		// console.log('Unrecognized signature version: %s', headers['box-signature-version']);
+		return undefined;
+	}
+
+	if (headers['box-signature-algorithm'] !== 'HmacSHA256') {
```
mself: The docs are wrong.  I'll get them fixed.

lib/managers/webhooks.js
mself: done

lib/managers/webhooks.js
mself: done

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+/**
+ * Compute the message signature (see https://docs.box.com/reference#signatures)
+ *
+ * @param {string} body - The request body of the webhook message
+ * @param {Object} headers - The request headers of the webhook message
+ * @param {string} signatureKey - The signature to verify the message with
+ * @returns {string} - The message signature
+ * @private
+ */
+function computeSignature(body, headers, signatureKey) {
```
mself: done

lib/managers/webhooks.js
```diff
@@ -193,6 +193,120 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+	if (!signatureKey) {
+		return undefined;
+	}
+
+	if (headers['box-signature-version'] !== '1') {
+		// console.log('Unrecognized signature version: %s', headers['box-signature-version']);
```
mself: I thought that they were useful to leave in.  If someone wants to extend or debug this code it will be much easier if they can comment these back in during development.  Is there a better way to do that?

lib/managers/webhooks.js
mself: done.  I added an additional (optional) parameter.

tests/lib/managers/webhooks-test.js
mself: done

lib/managers/webhooks.js
```diff
@@ -25,6 +26,90 @@ var urlPath = require('../util/url-path');
...
+// This prevents replay attacks
+const MAX_MESSAGE_AGE = 10 * 60; // 10 minutes
+
+/**
+ * Compute the message signature
+ * @see {@Link https://docs.box.com/reference#signatures}
+ *
+ * @param {string} body - The request body of the webhook message
+ * @param {Object} headers - The request headers of the webhook message
+ * @param {string} signatureKey - The signature to verify the message with
```
mattwiller: Based on the code, it looks like this paramter should be nullable: `{?string}`

lib/managers/webhooks.js
```diff
@@ -194,6 +279,30 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+ * @param {string} [secondarySignatureKey] - The secondary signature to verify the message with
+ * @param {int} [maxMessageAge] - The maximum message age (in seconds).  Defaults to 10 minutes
+ * @returns {boolean} - True or false
+ */
+Webhooks.validateMessage = function(body, headers, primarySignatureKey, secondarySignatureKey, maxMessageAge) {
+	if (!validateSignature(body, headers, primarySignatureKey, secondarySignatureKey)) {
+		return false;
+	}
+
+	if (!validateDeliveryTimestamp(headers, maxMessageAge | MAX_MESSAGE_AGE)) {
```
mattwiller: This is doing a bitwise OR, which probably isn't what you want here.  I would suggest doing something like `if (typeof maxMessageAge !== 'number')` at the top of the function and setting it to `MAX_MESSAGE_AGE` if not.

tests/lib/managers/webhooks-test.js
```diff
@@ -151,4 +152,73 @@ describe('Webhooks', function() {
...
+			const clock = sinon.useFakeTimers(DATE_IN_PAST);
+			assert.ok(!Webhooks.validateMessage(BODY, HEADERS_WITH_WRONG_SIGNATURE_VERSION, PRIMARY_SIGNATURE_KEY, SECONDARY_SIGNATURE_KEY));
+			clock.restore();
+		});
+
+		it('should NOT validate the webhook message if the signature algorithm is wrong', function() {
+			const clock = sinon.useFakeTimers(DATE_IN_PAST);
+			assert.ok(!Webhooks.validateMessage(BODY, HEADERS_WITH_WRONG_SIGNATURE_ALGORITHM, PRIMARY_SIGNATURE_KEY, SECONDARY_SIGNATURE_KEY));
+			clock.restore();
+		});
```
mattwiller: Could we add a test to ensure that the passed-in `maxMessageAge` value is used when it's passed in?

lib/managers/webhooks.js
mself: done

lib/managers/webhooks.js
mself: I used `(maxMessageAge === null) ? MAX_MESSAGE_AGE : maxMessageAge` since I didn't want a non-numeric value to get silently ignored (should result in an error instead).

mself: Ack!!!

done

tests/lib/managers/webhooks-test.js
mself: done

PR #110, opened 2017-02-27
Added support for streaming enterprise events

mself:


lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,204 @@
...
+// Public
+// ------------------------------------------------------------------------------
+
+/**
+ * Stream of Box enterprise events.
+ *
+ * By default, the stream starts from the current time.
+ * Pass 'startDate' to start from a specific time.
+ * Pass 'streamPosition' to start from a previous stream position, or '0' for all available past events (~1 year).
+ * If pollingInterval = 0, then the stream will end when all available events have been delivered (no polling).
```
mattwiller: This seems like a weird option, since there are already stream flow control options; why would a user want a stream that only makes one call?

README.md
```diff
@@ -194,6 +194,7 @@ The following resources are supported by the SDK:
 - [Device Pins](https://github.com/box/box-node-sdk/blob/master/docs/device-pins.md)
 - [Enterprise](https://github.com/box/box-node-sdk/blob/master/docs/enterprise.md)
 - [Events](https://github.com/box/box-node-sdk/blob/master/docs/events.md)
+- [Enterprise Events](https://github.com/box/box-node-sdk/blob/master/docs/enterprise-events.md)
```
mattwiller: Now that this is in the Events manager, this can be removed

docs/README.md
```diff
@@ -12,6 +12,7 @@ the SDK are available by topic:
 - [Device Pins](device-pins.md)
 - [Enterprise](enterprise.md)
 - [Events](events.md)
+- [Enterprise Events](enterprise-events.md)
```
mattwiller: Now that this is in the Events manager, this can be removed

docs/enterprise-events.md
```diff
@@ -0,0 +1,126 @@
+Enterprise Events
```
mattwiller: Now that this is in the Events manager, this should all be moved to `events.md`

docs/enterprise-events.md
```diff
@@ -0,0 +1,126 @@
...
+
+```js
+const stream = client.events.getEnterpriseEventStream({
+    startDate: '2016-01-01T00:00:00-08:00',
+    pollingInterval: 60
+});
+```
+
+Note that Box buffers enterprise events for ~60 seconds before making them available to the `/events` API
+(to ensure that events are delivered in-order and without duplicates), so polling with an inerval of less than
```
mattwiller: [typo] "inerval"

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,252 @@
...
+};
+
+/**
+ * Read and set the stream position and start date from a file.
+ *
+ * @param {string} file - A filename or file descriptor
+ * @param {Function} callback - Passed the stream position and start date
+ * @returns {void}
+ */
+EnterpriseEventStream.prototype.readStreamPositionFromFile = function(file, callback) {
```
mattwiller: This seems like the wrong level for the SDK; users can read and persist state however they want.  Why not expose a `getState()` method that returns the necessary state and a `resume()` method that takes the expected state object?  That way consumers could store state in Redis, etc. if they want

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,252 @@
...
+	this._client.events.get(params, function(err, result) {
+		if (err) {
+			self.emit('error', err);
+			// If there was a "permanent" error, we would call the callback with it here.
+			// But it's not clear which errors are truly permanent?
+			// If Box is down or returning errors for an extended period, we still want to resume when it recovers.
+			// So, continue polling at the regular frequency.
+			// Don't use a shorter retry interval (to avoid DDOSing Box).
+		} else if (result && result.entries) {
+			// console.log(`Fetched events from ${options.streamPosition || options.createdAfter} (limit ${options.limit}) => ${result.entries.length} events (next ${result.next_stream_position})`);
```
mattwiller: Please remove commented-out code before merging

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,252 @@
...
+				// Notify the reader of the new stream position.
+				// Stream readers can respond to the 'streamPosition' event to persist the stream position.
+				self.emit('streamPosition', self._streamPosition);
+
+				callback(null, result.entries);
+				return;
+			}
+		}
+
+		if (!self._options.pollingInterval) {
```
mattwiller: How does this work if the user wanted all events from one specific day in the past, but those events didn't all fit into one page from the API?  Or if I want all the events from the start of today until now?  I think there's potential for a mismatch between how a user might think of "polling" and how it's being used here; I would expect that this applies only to "future" events, not that I'll only ever get one page.

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,252 @@
...
+		// If polling is disabled, end the stream.
+			callback();
+			return;
+		}
+
+		// There were no events returned (or an error occurred), so schedule another poll.
+		const delay = self._options.pollingInterval * 1000;
+
+		// Stream readers can use this to flush buffered events to a downstream system.
+		self.emit('wait', delay);
```
mattwiller: I don't see this documented in the Stream docs; is this something Box-specific?  Can you elaborate on the use case for this?

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,252 @@
...
+			// Close the stream if there was a "permanent" failure or we reached the end of the events.
+			self.push(null);
+			return;
+		}
+
+		// Pause the stream to avoid race conditions while pushing in the new events.
+		// Without this, _read() would be called again from inside each push(),
+		// resulting in multiple parallel calls to fetchEvents().
+		// See https://github.com/nodejs/node/issues/3203
+		const wasPaused = self.isPaused();
```
mattwiller: I remember you mentioned this before; is this a problem for the User Events stream as well?

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,252 @@
...
+
+		// There were no events returned (or an error occurred), so schedule another poll.
+		const delay = self._options.pollingInterval * 1000;
+
+		// Stream readers can use this to flush buffered events to a downstream system.
+		self.emit('wait', delay);
+
+		setTimeout(function() {
+			self.fetchEvents(callback);
+		}, delay);
```
mattwiller: Per convention, I would flip the cases here; put this "error" case up near the top inside a conditional, and put the success case at the bottom.

lib/managers/events.js
```diff
@@ -156,4 +166,107 @@ Events.prototype.getEventStream = function(streamPosition, callback) {
...
+ * @param {string} [options.streamPosition] - The stream position to start from (pass '0' for all past events)
+ * @param {string} [options.startDate] - The date to start from
+ * @param {string} [options.endDate] - The date to end at
+ * @param {string} [options.eventTypeFilter] - Comma-separated list of event types to return
+ * @param {int} [options.pollingInterval=60] - Polling interval (in seconds).  Pass 0 for no polling.
+ * @param {int} [options.chunkSize=500] - Number of events to fetch per call (max = 500)
+ * @returns {EnterpriseEventStream} - The enterprise events stream
+ */
+Events.prototype.getEnterpriseEventStream = function(options) {
+	return new EnterpriseEventStream(this.client, options);
```
mattwiller: This method is synchronous, where the User Events stream is asynchronous.  Could we make them both follow the same convention?

lib/enterprise-event-stream.js
mself: Yes, I think it is.  Will come back to this once this is merged.

docs/events.md
```diff
@@ -1,6 +1,16 @@
...
 
+The Box API supports two types of event streams -- one for the events specific to a particular user
+ and one for all of the events in an enterprise.
+
+* [User Events](#user-events)
+* [Enterprise Events](#enterprise-events)
+
+
+User Events
+===========
```
mattwiller: The heading levels on these need to be adjusted; they should be second-level headings (under top-level "Events") using the `--------` underline.

docs/events.md
```diff
@@ -68,3 +78,129 @@ point in time:
...
+The SDK includes an `EnterpriseEventStream` class (which implements
+[stream.Readable](https://nodejs.org/api/stream.html#stream_readable_streams)) that automatically
+handles polling for events and delivering them to the application.
+
+* [Listening to the Enterprise Event Stream](#listening-to-the-enterprise-event-stream)
+* [Handling errors](#handling-errors)
+* [Get the Stream Position](#get-the-stream-position)
+
+Listening to the Enterprise Event Stream
+----------------------------------------
```
mattwiller: These heading levels also need to be adjusted to be third-level headings, using the `### Title` notation

docs/events.md
```diff
@@ -68,3 +78,129 @@ point in time:
...
+        // Reached the end of the stream.
+    });
+});
+```
+
+You can also filter the event stream to only receive specific event types.  The set of enterprise event types
+is available in `client.events.enterpriseEventTypes`.
+
+```js
+const stream = client.events.getEnterpriseEventStream({
```
mattwiller: This method no longer returns a stream, so this example needs to be updated

docs/events.md
```diff
@@ -68,3 +78,129 @@ point in time:
...
+client.events.getEnterpriseEventStream(function(err, stream) {
+	
+    if (err) { // Handle error }
+
+    // Restore the stream state from the previous run.
+    stream.setStreamState(readState());
+
+    stream.on('streamPosition', function() {
+        // Persist the stream state each time the stream position changes.
+        writeState(stream.getStreamState());
```
mattwiller: Why not have the streamPosition event pass the stream state as event data?

docs/events.md
```diff
@@ -68,3 +78,129 @@ point in time:
...
+same point if your application is interrupted and restarted.  You can attach a `'streamPosition'` event listener
+to be notified each time the stream position changes.
+
+```js
+client.events.getEnterpriseEventStream(function(err, stream) {
+	
+    if (err) { // Handle error }
+
+    // Restore the stream state from the previous run.
+    stream.setStreamState(readState());
```
mattwiller: It seems a little odd to get a stream and then set its state manually; perhaps we should add an `initialState` option to `getEnterpriseEventStream()` so the user can just pass in the state and have it set automatically.

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		// Only update the stream position if there were events returned.
+		// The API currently returns next_stream_position = 0 if there are no events (may be a bug?).
+		// But we don't want to start over at the beginning in that case, so ignore it.
+		self._streamPosition = result.next_stream_position;
+
+		// Notify the reader of the new stream position.
+		// Stream readers can respond to the 'streamPosition' event to persist the stream position.
+		self.emit('streamPosition', self._streamPosition);
```
mattwiller: It seems like it would be more useful to have a `'newStreamState'` event that send the full stream state object for persistence

tests/lib/managers/events-test.js
```diff
@@ -420,4 +428,33 @@ describe('Events', function() {
...
 		});
 	});
+
+	describe('getEnterpriseEventStream()', function() {
+
+		var options = {
+			streamPosition: TEST_STREAM_POSITION
+		};
+
+		it('should return event stream from starting stream position when passed stream position', function(done) {
```
mattwiller: This test name is incorrect; it should call the callback with the stream, not return it

lib/enterprise-event-stream.js
```diff
@@ -0,0 +1,234 @@
...
+ * Once the stream catches up to the current time, it will begin polling every 'pollingInterval' seconds.
+ * If 'pollingInterval' = 0, then the stream will end when it catches up to the current time (no polling).
+ *
+ * @param {BoxClient} client - The client to use to get events
+ * @param {Object} [options] - Options
+ * @param {string} [options.streamPosition] - The stream position to start from (pass '0' for all past events)
+ * @param {string} [options.startDate] - The date to start from
+ * @param {string} [options.endDate] - The date to end at
+ * @param {EventType[]} [options.eventTypeFilter] - Array of event types to return
+ * @param {int} [options.pollingInterval=60] - Polling interval (in seconds).  Pass 0 for no polling.
```
mattwiller: Is the `0` behavior here redundant with setting an endDate of the current timestamp?  Do we need both?

mself: Unfortunately, the API doesn't return anything to tell you that you have received all events up to `endDate`, so the SDK will keep polling indefinitely in this case.  Plus, endDate could be in the future.  Perhaps we can improve this if the API gets cleaner in the future?

docs/events.md
mself: Yes, I agree.  I think we can add this later without breaking anything.

PR #111, opened 2017-03-02
Support multi-put upload

mattwiller:
Putting this up for review; it's not quite ready to merge but the meat of the code is there so I'd like to start getting some eyes on it.

/cc @djordan


lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+/**
+ * Commit an upload session after all parts have been uploaded, creating the new file
+ *
+ * API Endpoint: '/files/upload-sessions/:sessionID/commit'
+ * Method: POST
+ *
+ * @param {string} sessionID - The ID of the upload session to commit
+ * @param {string} fileHash - The base64-encoded SHA-1 hash of the file being uploaded
+ * @param {UploadPart[]} parts - The list of uploaded chunks to be committed
+ * @param {?Object} options - Optional parameters, can be left null
```
KarthikShanmugasundaram: As attributes are file attributes (key value pairs). May be we could give a hint in the param description.

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+		this.stream = null;
+
+		this.client.files.abortUploadSession(this.sessionID, err => {
+
+			if (err) {
+				this.emit('abortFailed', err);
+				return;
+			}
+
+			this.emit('aborted');
```
bodawei: were it me, I'd be adding in a header comment to the class what the expected events are.  Searching through source code to try to figure out what events I should be listening to is a drag :-)

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+
+			buf = this.streamBuffer.shift();
+		} else {
+
+			// Stream case, need to read
+			buf = this.stream.read(this.partSize);
+
+			if (!buf) {
+				// stream needs to read more, retry later
+				setImmediate(() => this._getNextChunk(callback));
```
bodawei: why set immediate rather than setTimeout()?  (just curious)

lib/managers/files.js
```diff
@@ -4,23 +4,35 @@
 
 'use strict';
 
+// -----------------------------------------------------------------------------
+// Requirements
+// -----------------------------------------------------------------------------
 
-// ------------------------------------------------------------------------------
+/**
+ * @typedef {Object} UploadPart
```
mattwiller: In the hopefully near future, the API will pass us back this information directly and we'll pass it directly back to the user, so I think it's useful to retain this typedef over a constructor

lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+	}
+
+	var apiURL = this.client._uploadBaseURL + urlPath(BASE_PATH, UPLOAD_SESSION_SUBRESOURCE, sessionID),
+		partID = options.partID || generatePartID(),
+		params = {
+			headers: {
+				'Content-Type': 'application/octet-stream',
+				Digest: 'SHA=' + crypto.createHash('sha1').update(part).digest('base64'),
+				'X-Box-Part-Id': partID,
+				'Content-Range': `bytes ${offset}-${offset + part.length - 1}/${totalSize}`
```
mattwiller: Since the API URL is a simple concatenation instead of a more complicated interpolation, I thought it was slightly clearer to leave it using `+`

lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+		callback(null, {
+			part_id: partID,
+			offset: offset,
+			size: part.length
+		});
+	});
+};
+
+/**
+ * Commit an upload session after all parts have been uploaded, creating the new file
```
mattwiller: There will eventually be a full API documentation for it, and I've also added some relatively complete documentation for the SDK as well

lib/managers/files.js
mattwiller: Since the param is in the middle of the list, you have to pass _something_, but either `undefined` or `null` is acceptable here if you don't want to pass any options.

lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+ * @param {string} sessionID - The ID of the upload session to commit
+ * @param {string} fileHash - The base64-encoded SHA-1 hash of the file being uploaded
+ * @param {UploadPart[]} parts - The list of uploaded chunks to be committed
+ * @param {?Object} options - Optional parameters, can be left null
+ * @param {Function} callback - Passed the new file information if successful
+ * @returns {void}
+ */
+Files.prototype.commitUploadSession = function(sessionID, fileHash, parts, options, callback) {
+
+	var apiURL = this.client._uploadBaseURL + urlPath(BASE_PATH, UPLOAD_SESSION_SUBRESOURCE, sessionID, 'commit'),
```
mattwiller: Yeah, that's one of the weird parts of these endpoints.  We've never had to use the upload URL base with RESTful methods before, so I just did this to get it working.  This is something that's definitely ripe for a refactor at some point.

lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+
+	var self = this;
+	this.client.post(apiURL, params, function(err, response) {
+
+		if (err) {
+			callback(err);
+			return;
+		}
+
+		if (response.statusCode === 201) {
```
mattwiller: Our linter doesn't like switch statements with `return` in it very much (you have to also include `break`), so it doesn't end up looking much nicer that way

lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+
+/**
+ * Get a list of all parts that have been uploaded to an upload session
+ *
+ * API Endpoint: '/files/upload-session/:sessionID/parts'
+ * Method: GET
+ *
+ * @param {string} sessionID - The ID of the session to get a list of parts from
+ * @param {?Object} options - Optional parameters, can be left null
+ * @param {string} [options.marker] - Paging marker for the list of parts
```
mattwiller: Yeah, they should.  That's all documented in the [API docs](https://docs.box.com/reference#marker-based-paging)

lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+ * @param {int} [options.limit] - Maximum number of parts to return
+ * @param {Function} callback - Passed the list of parts if successful
+ * @returns {void}
+ */
+Files.prototype.getUploadSessionParts = function(sessionID, options, callback) {
```
mattwiller: Yeah, mostly just existing conventions keeping things all a single level deep from the client

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+		this.canceled = false;
+	}
+
+	/**
+	 * Get the final object representation of this chunk for the API
+	 * @returns {UploadPart} The chunk object
+	 */
+	getData() {
+
+		return this.data;
```
mattwiller: No real reason, I guess.  The way I was storing the data changed a few times during development, so I used a method to maintain a decent interface.

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+/** Manager for uploading a file in chunks */
+class UploadManager extends EventEmitter {
+
+	/**
+	 * Create an upload manager
+	 * @param {BoxClient} client The client to use to upload the file
+	 * @param {Object} uploadSessionInfo The upload session info to use for chunked upload
+	 * @param {ReadableStream|Buffer|string} file The file to upload
+	 * @param {int} size The size of the file to be uploaded
+	 * @param {Object} [options] Optional parameters
```
mattwiller: Since `options` is the last param here, it's truly optional instead of merely nullable, so I marked it as such.

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+	 */
+	constructor(client, uploadSessionInfo, file, size, options) {
+
+		super();
+
+		this.client = client;
```
mattwiller: Good point, I'll mark them all private

lib/upload-manager.js
mattwiller: We don't need to wait a specific amount of time, but rather just until after the next round of I/O occurs, so setImmediate should allow us to retry at the soonest reasonable time.

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+	 */
+	start() {
+
+		if (this.isStarted) {
+			return;
+		}
+
+		// Create the initial chunks
+		for (let i = 0; i < this.options.parallelism; i++) {
+			this._getNextChunk(chunk => (chunk ? this._uploadChunk(chunk) : this._commit()));
```
mattwiller: I think it should be fine, since the very last chunk to be uploaded must emit the `'uploaded'` event, which first decrements the number of chunks in flight to zero and then triggers the `_commit()` call.

lib/managers/files.js
```diff
@@ -968,6 +991,230 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
...
+/**
+ * Get a list of all parts that have been uploaded to an upload session
+ *
+ * API Endpoint: '/files/upload-session/:sessionID/parts'
+ * Method: GET
+ *
+ * @param {string} sessionID - The ID of the session to get a list of parts from
+ * @param {?Object} options - Optional parameters, can be left null
+ * @param {string} [options.marker] - Paging marker for the list of parts
+ * @param {int} [options.limit] - Maximum number of parts to return
```
mattwiller: I'm sure it does, but I don't know what it is yet since the public docs aren't available

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+			}
+
+			if (err) {
+				// handle the error or retry
+				if (err.statusCode) {
+					// an API error, probably not retryable!
+					this.emit('error', err);
+				} else {
+					// maybe a network error, retry
+					this.retry = setTimeout(() => this.upload(), RETRY_INTERVAL_MS);
```
mattwiller: Hmm, I'm not sure which way is best here, since ideally the uploader would just keep retrying until the connection comes back.  If the error doesn't have a status code, it didn't come from the API (at least not via HTTP)

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+	}
+}
+
+// -----------------------------------------------------------------------------
+// Public
+// -----------------------------------------------------------------------------
+
+//@TODO: rename to something better
+/** Manager for uploading a file in chunks */
+class UploadManager extends EventEmitter {
```
mattwiller: I didn't like the collision with our endpoint "managers", so I just dropped that part entirely

lib/upload-manager.js
```diff
@@ -0,0 +1,287 @@
...
+			buf = this.streamBuffer.shift();
+		} else {
+
+			// Stream case, need to read
+			buf = this.stream.read(this.partSize);
+
+			if (!buf) {
+				// stream needs to read more, retry later
+				setImmediate(() => this._getNextChunk(callback));
+				return;
```
mattwiller: It should never call commit until `_getNextChunk()` yields `null`; this just defers to a later event loop turn.

lib/chunked-uploader.js
```diff
@@ -16,21 +71,26 @@ const EventEmitter = require('events').EventEmitter,
 // Private
 // -----------------------------------------------------------------------------
 
-//@TODO: make into an option
+// @TODO: make into an option
```
djordan: Did you mean to leave this for later? I'm OK with that call, but just checking in case it was an oversight.

lib/managers/files.js
```diff
@@ -991,7 +992,7 @@ Files.prototype.deleteVersion = function(fileID, versionID, callback) {
 };
 
 /**
- * Creates an upload session to create a new file in chunks.  This will first
+ * Creates a session used to upload a new file in chunks..  This will first
```
djordan: Super nit: extra . at end of sentence

tests/integration-test.js
```diff
@@ -410,4 +411,282 @@ describe('Box Node SDK', function() {
...
+			)
+			.matchHeader('Authorization', function(authHeader) {
+				assert.equal(authHeader, 'Bearer ' + TEST_ACCESS_TOKEN);
+				return true;
+			})
+			.matchHeader('Digest', function(digestHeader) {
+				assert.equal(digestHeader, 'SHA=' + crypto.createHash('sha1').update(fileContents).digest('base64'));
+				return true;
+			})
+			.reply(201, file);
```
djordan: That is quite a chain of calls!

lib/managers/files.js
```diff
@@ -1071,81 +1067,51 @@ Files.prototype.commitUploadSession = function(sessionID, fileHash, options, cal
...
-		},
-		function commitUpload(waterfallCallback) {
+	return (userParts ? Promise.resolve(userParts) : fetchParts(0, []))
+		.then(parts => {
 
 			params.body.parts = parts;
-			self.client.post(apiURL, params, waterfallCallback);
-		}
-	], function(err, response) {
+			return this.client.post(apiURL, params);
```
djordan: It was kind of nice to have the named "commitUpload" function previously. Could you just add a comment here? e.g. // Commit upload

PR #116, opened 2017-04-07
can_view_path and notify support added

KarthikShanmugasundaram:


lib/managers/collaborations.js
```diff
@@ -119,7 +119,7 @@ Collaborations.prototype.respondToPending = function(collaborationID, newStatus,
  * @param {string} itemID - Box ID of the item to which the user should be invited
  * @param {CollaborationRole} role - The role which the invited collaborator should have
  * @param {Object} [options] - Optional parameters for the collaboration
- * @param {ItemType} [options.type=folder] - Type of object to be collaborated
+ * @param {ItemType} [options.type=folder] or [options.type=file]  - Type of object to be collaborated
```
mattwiller: That's not how this JSDoc is used; the type `{ItemType}` gives all available values.  The [options.type=folder]` is just saying that the parameter is optional with a default value of `"folder"`.  Also, please add parameter descriptions for the new options you're adding!

lib/managers/collaborations.js
```diff
@@ -146,6 +146,17 @@ Collaborations.prototype.create = function(accessibleBy, itemID, role, options,
...
+
+	if (options.can_view_path) {
+		params.body.can_view_path = options.can_view_path;
+	}
+
+	if (options.notify) {
+		params.qs = {
+			notify: options.notify
+		};
+	}
```
mattwiller: For both of these parameters, a value of `false` will NOT be passed (since it fails the if check).  You'll need to do e.g. a `typeof options.notify === 'boolean'` check instead of simple loose truthiness.

lib/managers/collaborations.js
```diff
@@ -159,7 +170,7 @@ Collaborations.prototype.create = function(accessibleBy, itemID, role, options,
  * @param {string} itemID - Box ID of the item to which the user should be invited
  * @param {CollaborationRole} role - The role which the invited collaborator should have
  * @param {Object} [options] - Optional parameters for the collaboration
- * @param {ItemType} [options.type=folder] - Type of object to be collaborated
+ * @param {ItemType} [options.type=folder] or [options.type=file] - Type of object to be collaborated
```
mattwiller: Not the correct use of JSDoc (see above) — please remove

lib/managers/collaborations.js
```diff
@@ -187,7 +198,7 @@ Collaborations.prototype.createWithUserID = function(userID, itemID, role, optio
...
```
mattwiller: Not the correct use of JSDoc (see above) — please remove

lib/managers/collaborations.js
```diff
@@ -215,7 +226,7 @@ Collaborations.prototype.createWithUserEmail = function(email, itemID, role, opt
  * @param {string} itemID - Box ID of the item to which the group should be invited
  * @param {CollaborationRole} role - The role which the invited collaborator should have
  * @param {Object} [options] - Optional parameters for the collaboration
- * @param {ItemType} [options.type=folder] - Type of object to be collaborated
+ * @param {ItemType} [options.type=folder] or [options.type=file] - Type of object to be collaborated
```
mattwiller: Not the correct use of JSDoc (see above) — please remove

tests/endpoint-test.js
```diff
@@ -283,6 +283,45 @@ describe('Endpoint', function() {
 				});
 			});
 
+			it('should make correct request with additional parameters and correctly parse response when API call is successful', function(done) {
```
mattwiller: Endpoint tests are generally expensive, so we only do one happy path test per endpoint.  Tests like these should be unit tests instead

tests/lib/managers/collaborations-test.js
```diff
@@ -366,6 +426,39 @@ describe('Collaborations', function() {
...
+			expectedParams.body.can_view_path = true;
+
+			expectedParams.qs = {
+				notify: true
+			};
+
+			var params = {
+				type: 'file',
+				notify: true,
+				can_view_path: true
```
mattwiller: Please add tests for the false case as well

KarthikShanmugasundaram: done

tests/endpoint-test.js
KarthikShanmugasundaram: removed

lib/managers/collaborations.js
KarthikShanmugasundaram: All the jsdoc changes done

PR #118, opened 2017-04-07
upload file by path implementation

KarthikShanmugasundaram:


lib/managers/files.js
```diff
@@ -541,6 +542,26 @@ Files.prototype.uploadFile = function(parentFolderID, filename, content, callbac
...
+ * @param {string} parentFolderID - the id of the parent folder to upload to
+ * @param {string} filename - the file name that the uploaded file should have
+ * @param {string} path - the full path of the file that needs to be uploaded. The path includes the filename
+ * @param {Function} callback - called with data about the upload if successful, or an error if the
+ * upload failed
+ * @returns {void}
+ */
+Files.prototype.uploadFileFromPath = function(parentFolderID, filename, path, callback) {
+
+	var stream = fs.createReadStream(path);
```
mattwiller: You should handle any errors thrown here!  This method is supposed to be asynchronous, so any errors produced should be passed to the callback rather than thrown.

tests/lib/managers/files-test.js
```diff
@@ -666,6 +666,57 @@ describe('Files', function() {
...
 	});
 
+	describe('uploadFileFromPath()', function() {
+
+		var PARENT_FOLDER_ID = '123',
+			FILENAME = 'folders-test.js',
+			PATH = './folders-test.js';
+
+		var streamFake = {on: sandbox.stub()};
+		it('should call BoxClient.upload() with the correct non-callback params', function() {
```
mattwiller: This test is not even testing the method you say it is!  You're calling files.uploadFile() in the test, not files.uploadFileFromPath()

tests/lib/managers/files-test.js
```diff
@@ -698,6 +749,41 @@ describe('Files', function() {
 		});
 	});
 
+	describe('uploadNewFileVersionFromPath()', function() {
+
+		var	PATH = './folders-test.js';
+
+		it('should call BoxClient.upload() with the correct non-callback params', function() {
+			sandbox.stub(boxClientFake, 'defaultResponseHandler');
+			sandbox.mock(boxClientFake).expects('upload').withArgs('/files/1234/content', null);
```
mattwiller: This is not the correct arguments that we expect to be passed to this mock!  It should not get null for the file content!

lib/managers/files.js
KarthikShanmugasundaram: done.

tests/lib/managers/files-test.js
KarthikShanmugasundaram: corrected.

tests/lib/managers/files-test.js
KarthikShanmugasundaram: done.

PR #119, opened 2017-04-07
Update metadata builder implementation

KarthikShanmugasundaram:


lib/util/update-metadata-builder.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/**
+ * Insert add operation to the request body.
+ *
+ * @param {string} path - path designates the key. it must be prefixed with /.
+ * @param {object} value - the value could any one of the types string, enum, float, and date (RFC 3339).
+ * @returns {void}
+ */
+UpdateMetadataBuilder.prototype.insertAdd = function(path, value) {
```
mattwiller: Since all these begin with 'insert', maybe it's easier to just name them after the operation, e.g. "add" or "replace"

lib/util/update-metadata-builder.js
```diff
@@ -0,0 +1,149 @@
...
+ * Insert path based operation in to the request body.
+ *
+ * @param {String} op - The possible operations are add, replace, test.
+ * @param {string} path - path designates the key. it must be prefixed with /.
+ * @param {object} value - the value could any one of the types string, enum, float, and date (RFC 3339).
+ * @returns {void}
+ */
+UpdateMetadataBuilder.prototype.insertPath = function(op, path, value) {
+	assert(typeof op === 'string' && op.length > 0, 'op designates the operation. It is mandatory.');
+	assert(typeof path === 'string' && path.length > 0, 'path designates the key. It is mandatory.');
```
mattwiller: While you're checking this, it's probably a good idea to ensure the path starts with '/'

lib/util/update-metadata-builder.js
```diff
@@ -0,0 +1,149 @@
...
+	};
+	this.body.push(item);
+};
+
+/**
+ * Get the request body. It contains array of opertions.
+ *
+ * @returns {array} the array of operations.
+ */
+UpdateMetadataBuilder.prototype.getRequestBody = function() {
```
mattwiller: "RequestBody" seems like strange language here, since this class has nothing to do with an API request.  Maybe just call it "getPatch"?

lib/util/update-metadata-builder.js
```diff
@@ -0,0 +1,149 @@
...
+
+var assert = require('assert');
+
+/**
+ * Utility for building the metadata patch request for files and folders
+ *
+ * @constructor
+ * @returns {void}
+ */
+function UpdateMetadataBuilder() {
```
mattwiller: The word order here is a little odd, since "UpdateMetadata" isn't a noun that can be built.  Consider changing to "MetadataUpdateBuilder"

tests/lib/util/update-metadata-builder-test.js
```diff
@@ -0,0 +1,155 @@
...
+		});
+
+		it('should throw when path is not a string', function() {
+			assert.throws(function() {
+				builder.insertRemove(10);
+			});
+		});
+	});
+
+	describe('methods()', function() {
```
mattwiller: `methods()` isn't a function on the class; this should be renamed

tests/lib/util/update-metadata-builder-test.js
```diff
@@ -0,0 +1,155 @@
...
+		});
+
+		it('should throw when value is undefined', function() {
+			assert.throws(function() {
+				builder.insertPath('add', 10);
+			});
+		});
+	});
+
+	describe('insertFrom()', function() {
```
mattwiller: These tests only test the failure cases; we should also have unit tests to ensure that the methods do what they should in the happy path.

lib/util/update-metadata-builder.js
KarthikShanmugasundaram: done

lib/util/update-metadata-builder.js
KarthikShanmugasundaram: done

lib/util/update-metadata-builder.js
KarthikShanmugasundaram: done

lib/util/update-metadata-builder.js
KarthikShanmugasundaram: done

tests/lib/util/update-metadata-builder-test.js
KarthikShanmugasundaram: done

tests/lib/util/update-metadata-builder-test.js
KarthikShanmugasundaram: done

PR #121, opened 2017-04-14
Box app settings implementation

KarthikShanmugasundaram:


lib/box-node-sdk.js
```diff
@@ -117,6 +117,36 @@ BoxSDKNode.accessLevels = BoxSDKNode.prototype.accessLevels = BoxClient.prototyp
 BoxSDKNode.collaborationRoles = BoxSDKNode.prototype.collaborationRoles = BoxClient.prototype.collaborationRoles;
 BoxSDKNode.CURRENT_USER_ID = BoxSDKNode.prototype.CURRENT_USER_ID = BoxClient.prototype.CURRENT_USER_ID;
 
+BoxSDKNode.getInstance = function(appConfig) {
```
mattwiller: I would add something to this method name to indicate what it's used for; this isn't really the "primary" way to get an instance, since it requires a special config object from the Dev Console

lib/box-node-sdk.js
```diff
@@ -117,6 +117,36 @@ BoxSDKNode.accessLevels = BoxSDKNode.prototype.accessLevels = BoxClient.prototyp
...
+
+	//Assign publicKeyID to keyID
+	if (typeof boxAppSettings.appAuth === 'object') {
+		var keyID = boxAppSettings.appAuth.publicKeyID;
+		if (typeof keyID === 'string') {
+			boxAppSettings.appAuth.keyID = keyID;
+		}
+	}
+
+	return new BoxSDKNode(boxAppSettings);
```
mattwiller: Rather than passing this object through directly, it might be better to construct our own object by pulling things off of `boxAppSettings`, just to make sure no random properties make it through that we don't expect.

lib/managers/webhooks.js
```diff
@@ -289,6 +289,14 @@ Webhooks.prototype.delete = function(webhookID, callback) {
  * @returns {boolean} - True or false
  */
 Webhooks.validateMessage = function(body, headers, primarySignatureKey, secondarySignatureKey, maxMessageAge) {
+	if (!primarySignatureKey && Webhooks.primaryKey) {
```
mattwiller: We should add these properties to the Webhooks object at the top of the file (but set them to `null` and add JSDoc for them so it's more obvious what they're doing.

lib/util/config.js
```diff
@@ -59,6 +59,8 @@ var defaults = {
 	expiredBufferMS: 30000,
 	staleBufferMS: 180000,
 	appAuth: undefined,
+	webhooks: undefined,
```
mattwiller: It doesn't look like the webhooks config will ultimately be used from here; do we really need to add it?

tests/lib/box-sdk-node-test.js
```diff
@@ -131,6 +151,36 @@ describe('box-node-sdk', function() {
 		});
 	});
 
+	describe.only('getInstance()', function() {
```
mattwiller: You need to remove the `.only()` before merging

lib/managers/webhooks.js
KarthikShanmugasundaram: @mattwiller I need to talk to you about the approach

lib/util/config.js
KarthikShanmugasundaram: done

lib/managers/webhooks.js
KarthikShanmugasundaram: done

lib/box-node-sdk.js
KarthikShanmugasundaram: Strict validation means, some of the custom properties such apiRootURL etc. is not part of the app settings. I like to discuss this with you.

KarthikShanmugasundaram: add configure() method as discussed.

lib/box-node-sdk.js
```diff
@@ -118,6 +118,67 @@ BoxSDKNode.collaborationRoles = BoxSDKNode.prototype.collaborationRoles = BoxCli
 BoxSDKNode.CURRENT_USER_ID = BoxSDKNode.prototype.CURRENT_USER_ID = BoxClient.prototype.CURRENT_USER_ID;
 
 /**
+ * Gets the BoxSDKNode instance by passing boxAppSettings json downloaded from the developer console.
+ *
+ * @param {object} appConfig boxAppSettings json is passed as parameter.
```
mattwiller: A couple issues with this JSDoc:
1. Non-primitive types should de documented capitalized, e.g. `{Object}`
1. The parameter description is a bit confusing; this takes an Object, not a JSON string
1. No need to say it's passed as a parameter, since this is a `@param` tag after all

lib/box-node-sdk.js
```diff
@@ -118,6 +118,67 @@ BoxSDKNode.collaborationRoles = BoxSDKNode.prototype.collaborationRoles = BoxCli
...
+ * @returns {BoxSDKNode} the BoxSDKNode instance
```
mattwiller: It might be good to describe the return value better; WHICH instance?  It's an instance that has been preconfigured with the values from the Dev Console

lib/box-node-sdk.js
```diff
@@ -118,6 +118,67 @@ BoxSDKNode.collaborationRoles = BoxSDKNode.prototype.collaborationRoles = BoxCli
...
+		}
+
+		appAuth.privateKey = boxAppSettings.appAuth.privateKey;
+
+		var passphrase = boxAppSettings.appAuth.passphrase;
+		if (typeof passphrase === 'string') {
+			appAuth.passphrase = passphrase;
+		}
+	}
+	params.appAuth = appAuth;
```
mattwiller: This is a little wonky with appAuth defined inside the block above.  I think it works due to hoisting, but we should be more explicit and set it to `null` or `undefined` above the `if` block to make it more clear what this does when the `if` case is not run.

lib/box-node-sdk.js
```diff
@@ -118,6 +118,67 @@ BoxSDKNode.collaborationRoles = BoxSDKNode.prototype.collaborationRoles = BoxCli
...
+
+	if (typeof boxAppSettings.enterpriseID === 'string') {
+		params.enterpriseID = boxAppSettings.enterpriseID;
+	}
+
+	return new BoxSDKNode(params);
+};
+
+/**
+ * Configures additional properties passed as parameter.
```
mattwiller: This description is a bit hard to understand.  Perhaps something like "Updates the SDK configuration with new parameters" would be clearer

lib/box-node-sdk.js
```diff
@@ -171,6 +232,14 @@ BoxSDKNode.prototype.getAnonymousClient = function() {
  * @returns {BoxClient} A new client authorized as the app user or enterprise
  */
 BoxSDKNode.prototype.getAppAuthClient = function(type, id) {
+	if (type === 'enterprise' && !id) {
+		if (this.config.enterpriseID) {
+			id = this.config.enterpriseID;
+		} else {
+			throw new Error('Enterprise id must be passed');
```
mattwiller: `id` should be capitalized

lib/box-node-sdk.js
```diff
@@ -219,6 +288,14 @@ BoxSDKNode.prototype.getTokensRefreshGrant = function(refreshToken, callback) {
  * @returns {void}
  */
 BoxSDKNode.prototype.getEnterpriseAppAuthTokens = function(enterpriseID, callback) {
+	if (!enterpriseID) {
+		if (this.config.enterpriseID) {
+			enterpriseID = this.config.enterpriseID;
+		} else {
+			throw new Error('Enterprise id must be passed');
```
mattwiller: `id` should be capitalized

lib/managers/legal-hold-policies.js
```diff
@@ -64,6 +64,7 @@ LegalHoldPolicies.prototype.assignmentTypes = Object.freeze({
  * @param {string} [options.description] - Description of the legal hold policy
  * @param {string} [options.filter_started_at] - Date filter, any Custodian assignments will apply only to file versions created or uploaded inside of the date range
  * @param {string} [options.filter_ended_at] - Date filter, any Custodian assignments will apply only to file versions created or uploaded inside of the date range
+ * @param {boolean} [options.is_ongoing] - After initialization, Assignments under this Policy will continue applying to files based on events, indefinitely
```
mattwiller: This isn't related to the rest of the change; please don't mix together separate changes if possible.

lib/managers/webhooks.js
```diff
@@ -126,6 +126,19 @@ function Webhooks(client) {
 }
 
 /**
+ * Primary signature key to protect webhooks against attacks.
+ * @static
```
mattwiller: These two JSDoc blocks should also have type information, e.g. `@type {?string}`

lib/managers/webhooks.js
```diff
@@ -279,21 +292,46 @@ Webhooks.prototype.delete = function(webhookID, callback) {
...
+};
+
+/**
  * Validate a webhook message by verifying the signature and the delivery timestamp
  *
  * @param {string} body - The request body of the webhook message
  * @param {Object} headers - The request headers of the webhook message
- * @param {string} primarySignatureKey - The primary signature to verify the message with
- * @param {string} [secondarySignatureKey] - The secondary signature to verify the message with
+ * @param {string} primaryKey - The primary signature to verify the message with. If it is sent as a parameter,
```
mattwiller: This parameter is now optional as well

tests/lib/managers/webhooks-test.js
```diff
@@ -153,6 +153,26 @@ describe('Webhooks', function() {
 
 	});
 
+	describe('setSignatureKeys()', function() {
+		const PRIMARY_SIGNATURE_KEY = 'SamplePrimaryKey';
+		const SECONDARY_SIGNATURE_KEY = 'SampleSecondaryKey';
+
+		it('should validate the primary signature key is set', function() {
```
mattwiller: This test name isn't very descriptive; maybe something like "should set the primary and secondary keys when both are passed"?

tests/lib/managers/webhooks-test.js
```diff
@@ -153,6 +153,26 @@ describe('Webhooks', function() {
...
+			Webhooks.setSignatureKeys(PRIMARY_SIGNATURE_KEY, SECONDARY_SIGNATURE_KEY);
+
+			assert.equal(PRIMARY_SIGNATURE_KEY, Webhooks.primarySignatureKey);
+			assert.equal(SECONDARY_SIGNATURE_KEY, Webhooks.secondarySignatureKey);
+		});
+
+		it('should validate the secondary signature key is undefined', function() {
```
mattwiller: This test name isn't very descriptive; maybe something like "should set the primary key when no secondary key is passed"?

tests/lib/managers/webhooks-test.js
```diff
@@ -153,6 +153,26 @@ describe('Webhooks', function() {
...
 	});
 
+	describe('setSignatureKeys()', function() {
+		const PRIMARY_SIGNATURE_KEY = 'SamplePrimaryKey';
+		const SECONDARY_SIGNATURE_KEY = 'SampleSecondaryKey';
+
+		it('should validate the primary signature key is set', function() {
+			Webhooks.setSignatureKeys(PRIMARY_SIGNATURE_KEY, SECONDARY_SIGNATURE_KEY);
+
+			assert.equal(PRIMARY_SIGNATURE_KEY, Webhooks.primarySignatureKey);
```
mattwiller: The order of the `assert.equal()` parameters in these tests is reversed.  Chai takes the actual value first, and the expected value second (see http://chaijs.com/api/assert/#method_equal).  This is important for error messaging when the tests fail.

tests/lib/box-sdk-node-test.js
```diff
@@ -131,6 +160,56 @@ describe('box-node-sdk', function() {
...
+		it('should verify that additional parameters can be passed to the BoxSDKNode instance', function() {
+			var additonalParams = {
+				apiRootURL: 'myUrl',
+				retryIntervalMS: 11111,
+				numMaxRetries: 3
+			};
+
+			sdk.configure(additonalParams);
+			assert.equal(sdk.config.clientID, 'myId');
+			assert.equal(sdk.config.apiRootURL, 'myUrl');
```
mattwiller: Assertions like this are easier if you use something like `assert.deepPropertyVal(sdk, 'config.apiRootURL', 'myUrl')`, which provides better error messaging if any of the intermediate objects is not present.

lib/box-node-sdk.js
KarthikShanmugasundaram: done

lib/box-node-sdk.js
KarthikShanmugasundaram: done

lib/box-node-sdk.js
KarthikShanmugasundaram: done

lib/box-node-sdk.js
KarthikShanmugasundaram: done

lib/box-node-sdk.js
KarthikShanmugasundaram: done

lib/box-node-sdk.js
KarthikShanmugasundaram: done

lib/managers/legal-hold-policies.js
KarthikShanmugasundaram: done

lib/managers/webhooks.js
KarthikShanmugasundaram: done

lib/managers/webhooks.js
KarthikShanmugasundaram: done

tests/lib/managers/webhooks-test.js
KarthikShanmugasundaram: done

tests/lib/managers/webhooks-test.js
KarthikShanmugasundaram: done

tests/lib/managers/webhooks-test.js
KarthikShanmugasundaram: done

tests/lib/box-sdk-node-test.js
KarthikShanmugasundaram: done

lib/box-node-sdk.js
```diff
@@ -167,11 +228,19 @@ BoxSDKNode.prototype.getAnonymousClient = function() {
  * Users themselves (as a user).
  *
  * @param {string} type The type of entity to operate as, "enterprise" or "user"
- * @param {string} id The Box ID of the entity to operate as
+ * @param {string} ID The Box ID of the entity to operate as
```
mattwiller: Variable names follow a camel-case convention, so this should be `id`

lib/box-node-sdk.js
```diff
@@ -167,11 +228,19 @@ BoxSDKNode.prototype.getAnonymousClient = function() {
...
  * @returns {BoxClient} A new client authorized as the app user or enterprise
  */
-BoxSDKNode.prototype.getAppAuthClient = function(type, id) {
-	var appAuthSession = new AppAuthSession(type, id, this.config, this.tokenManager);
+BoxSDKNode.prototype.getAppAuthClient = function(type, ID) {
+	if (type === 'enterprise' && !ID) {
+		if (this.config.enterpriseID) {
+			ID = this.config.enterpriseID;
+		} else {
+			throw new Error('Enterprise id must be passed');
```
mattwiller: In text, we prefer the usual "ID"

lib/box-node-sdk.js
mattwiller: This should still be capitalized

mself:
@KarthikShanmugasundaram Can you check if it is possible to create a new BoxSDK object using a JSON config that includes webhook info (and client ID), but which does NOT include an appAuth section (no private key, etc.).  This case happens when you create a client for the purpose of receiving webhooks but you aren't making any API calls from that client.  Currently, I think this might throw an exception since the private key is missing.

KarthikShanmugasundaram:
@mself This is not a problem witht the latest code.

PR #122, opened 2017-04-19
Retry JWT grant with Box server time when exp claim fails

### LINKED ISSUE #109: Intermittent "auth code or refresh token expired"/"Expired Auth: Auth code or refresh token has expired" error with App Auth client at v1.3.0
This looks like the same problem as #96, but I've seen it at v1.2.0 and v1.3.0.

This is the code that is producing the error:

```Javascript
 sdk.folders.getItems(req.params.folderId, null, (err, data) => {
    if (err) {
      console.log(`An error occurred - ${err.message}`);
      return res.status(500).send(err);
    }
    console.log(`All good -  ${data.id}`);
    return res.status(200).json(data);
  });
```

This is the output I get on the console:
```
An error occurred - Expired Auth: Auth code or refresh token has expired.
```

The client that is receiveing the response from `res.status(500).send(err)` receives this (certain bits omitted here for brevity):
```json
{
  "statusCode":400,
  "response":{
    "statusCode":400,
    "body":{
      "error":"invalid_grant",
      "error_description":"Current date/time MUST be before the expiration date/time listed in the 'exp' claim"
    },
    "headers": ...,
    "request":{
      "uri":{
        ...
        "href":"https://api.box.com/oauth2/token"
      },
      "method":"POST",
      "headers": ...
    }
  },
  "authExpired":true
}
```

To start with, it appeared that restarting node stopped the error occurring for a while. But now it's occurring all the time.

### LINKED ISSUE #115: Feature request: Use server timestamp to generate exp claim
I would like to request a feature (or to ask for input on whether a pull request for this feature would be likely to be accepted). The feature would be this: when a token request fails with the error "`Current date/time MUST be before the expiration date/time listed in the 'exp' claim`", then the SDK would take the "Date" header form the response, and generate a new `exp` claim using the Date header's value as the current timestamp, in an effort to generate an `exp` claim that is correct relative to the Box server's timestamp.

This is in an attempt to avoid the problem discussed in #109 when the clocks are not synchronised between the client machine and Box's server.

One downside of this is that the error keyword (`invalid_grant`) is not specific to the `exp` claim. So either the SDK would have to do this for all `invalid_grant` errors (or at least it would check for mismatched timestamps whenever it receives an `invalid_grant` error & only retry if it finds they're out of sync), or it would have to inspect the `error_description`, which is probably not a reliable thing to do.

(There are alternative approaches that could be taken to address that problem, such as moving the generation of the `exp` claim entirely onto Box's server, but that would be a more fundamental change in the API.)

Is this something that would make sense in this project, or would it be out of place?

mattwiller:
Fixes #115
Fixes #109

When a client's clock is desynchronized from Box API servers, JWT
grant calls might fail due to the exp claim not being valid.  In
order to mitigate this type of error, the SDK will attempt one
retry using the timestamp sent back in the Date header of the API
error response, which should correct most cases of this error.

### DIFF
--- a/lib/token-manager.js
+++ b/lib/token-manager.js
@@ -341,7 +341,7 @@ TokenManager.prototype = {
 		null,
 		function(err, tokenInfo) {
 
-            // When a client's clock is out of sync with Box API servers, they'll get an error about the exp claim
+			// When a client's clock is out of sync with Box API servers, they'll get an error about the exp claim
 			// In these cases, we can attempt to retry the grant request with a new exp claim calculated frem the
 			// Date header sent by the server
 			if (err && err.authExpired && err.response.body.error_description.indexOf('exp') > -1 && err.response.headers.date) {
--- a/lib/token-manager.js
+++ b/lib/token-manager.js
@@ -306,6 +306,7 @@ TokenManager.prototype = {
 		}
 
 		var claims = {
+			exp: Math.floor(Date.now() / 1000) + this.config.appAuth.expirationTime,
 			box_sub_type: type
 		};
 		var options = {
@@ -314,7 +315,6 @@ TokenManager.prototype = {
 			subject: id,
 			issuer: this.config.clientID,
 			jwtid: uuid.v4(),
-			expiresIn: this.config.appAuth.expirationTime,
 			noTimestamp: !this.config.appAuth.verifyTimestamp,
 			headers: {
 				kid: this.config.appAuth.keyID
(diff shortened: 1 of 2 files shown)


MisterS:
Hi! When are you going to release this over NPM? We urgently need this fix, as we constantly struggle with this.

mattwiller:
@MisterS We're targeting a release within the next couple days!  It's worth noting though that the root cause of this issue is that the clock of the machine you're running on is not synchronized correctly to UTC; you could also fix this issue by ensuring that your machine has the correct time, e.g. by using ntp.

MisterS:
Hi Matt!
I am aware of that cause, but somehow we are not able to keep the time in sync - don't ask me why. We synched our (hosted) webserver several times - by windows time server and manually - but after a short time the synching seems to be off again and the request break again. Though we have the problem, even if our server timestamp differs only around 3 seconds from the time of the unixtimestamp.com server.
Any ideas on this? Atm we are a little desperate here ...

mattwiller:
@MisterS Sorry for the delayed response — I was out of office without access to a computer.  We'll be releasing a new version of the SDK with this fix included in the next couple days!

mattwiller:
@MisterS Just to close the loop on this, you can now get v1.4.1 of the SDK on npm, which contains the fix for this issue.

MisterS:
Hi Matt, thank you very much for your info. In the meantime we helped us with the "noTimestamp" flag. But we will reconsider this option, as it is more safe. Appreciate your support!

ksleima:
@mattwiller I have tried the 1.5 sdk and this is still an issue. keep getting error: 'invalid_grant', error_description: 'Please check the \'exp\' claim. Are you sure this works?

nimeshvaghasiya:
hi @mattwiller
I am facing the same issue as @ksleima

I'm using box SDK version `"box-node-sdk": "^1.5.1"`
I'm attempting login and get above error.

error header is: 
`headers: { server: 'ATS', date: 'Fri, 23 Jun 2017 19:06:26 GMT', 'content-type': 'application/json', 'content-length': '77', 'cache-control': 'no-store', 'strict-transport-security': 'max-age=31536000; includeSubDomains', vary: 'Accept-Encoding', 'set-cookie': [Object], age: '0', via: 'https/1.1 adc-traffic-manager1015 (ApacheTrafficServer/6.2.1), https/1.1 ss-3av8 (ApacheTrafficServer/6.2.1)', connection: 'keep-alive' }`

When I send login request my Windows PC time was 'Sat, 24 Jun 2017 00:36:26'

also, I have used below options but not worked for me.
`expirationTime: 120,
noTimestamp:true`

When i not set expirationTime: 120 option i got 500 internal server error.

regards,
nimesh

nimeshvaghasiya:
error header shows below error.
`body: { error: 'invalid_grant', error_description: 'Please check the \'exp\' claim.' } }, authExpired: true`

mattwiller:
Hi @nimeshvaghasiya — the Box API does not allow expiration times more than 60 seconds in the future, so setting `expirationTime` to 120 will cause all requests to fail.  Please try lowering `expirationTime` to 30 and let me know if that fixes the errors!

nimeshvaghasiya:
@mattwiller I have upgraded to version 1.6.0. And it works fine without setting expirationTime.
Thanks.

PR #123, opened 2017-04-20
Add iterator support for collections

mattwiller:
To help users page through collections, this adds an option to the
SDK to return iterators from the default response handler for
collection responses (i.e. GET requests that return an object with
the entries field).  The iterator conforms to the proposed spec
for async iterators, which is currently nearing the final stages
for being accepted into a future ES version. See
https://github.com/tc39/proposal-async-iteration for more details.

This approach maximizes future compatibility (e.g. with upcoming
syntax like for-await-of loops), while minimizing the surface area
of the change and keeping it backwards-compatible.


lib/util/paging-iterator.js
```diff
@@ -0,0 +1,190 @@
...
+	 * @param {Object} response - The original API response
+	 * @param {BoxClient} client - An API client to make further requests
+	 * @returns {void}
+	 * @throws {Error} Will throw when collection cannot be paged
+	 */
+	constructor(response, client) {
+
+		var data = response.body;
+
+		if (!Array.isArray(data.entries)) {
```
djordan: Maybe the constructor should check isIterable() as well?

lib/util/paging-iterator.js
```diff
@@ -0,0 +1,190 @@
...
+		}));
+	}
+
+	/**
+	 * Fetch the next page of the collection
+	 * @returns {Promise} Promise resolving to iterator state
+	 */
+	next() {
+
+		if (this.nextChunk) {
```
djordan: I'm a little confused about this.nextChunk since you only use it here and in the constructor. Is it only for the initial results?

lib/util/paging-iterator.js
```diff
@@ -0,0 +1,190 @@
...
+		this.done = false;
+
+		var queryIndex = response.request.href.indexOf('?');
+		var href = queryIndex > 0 ? response.request.href.substring(0, queryIndex) : response.request.href;
+		this.options = {
+			headers: response.request.headers,
+			qs: querystring.parse(response.request.uri.query)
+		};
+
+		// qs.parse() makes everything a string, ensure the limit is numeric
```
mattwiller: There could be, but it doesn't matter because we'll just reserialize them (querystring doesn't differentiate between numeric and string types).  The reason we need to ensure limit is numeric is so when we do math with it we get number math and not string concatenation.

lib/util/paging-iterator.js
mattwiller: Yes, it is — although if we decide to iterate over individual items instead of whole pages it might serve a broader purpose

lib/util/paging-iterator.js
```diff
@@ -0,0 +1,190 @@
...
+			this.nextValue += this.limit;
+
+			if (Number.isSafeInteger(data.total_count)) {
+				this.done = data.offset + this.limit >= data.total_count;
+			} else {
+				this.done = data.entries.length === 0;
+			}
+		} else if (this.nextField === PAGING_MODES.MARKER) {
+
+			if (data.next_marker) {
```
mattwiller: Yep, `undefined`, `null`, and `""` are all falsy (but e.g. '0' is truthy)

lib/util/paging-iterator.js
```diff
@@ -0,0 +1,190 @@
...
+	 * @param {Object} response - The latest API response
+	 * @returns {void}
+	 */
+	_updatePaging(response) {
+
+		var data = response.body;
+
+		if (this.nextField === PAGING_MODES.OFFSET) {
+
+			this.nextValue += this.limit;
```
mattwiller: This already uses the limit returned by the API (or the length or the `entries` array if the API didn't return a `limit` field; see the constructor for details

PR #126, opened 2017-04-28
if id is not passed, getAppAuthClient should throw an error

KarthikShanmugasundaram:


tests/lib/box-sdk-node-test.js
```diff
@@ -286,6 +286,14 @@ describe('box-node-sdk', function() {
 			assert.ok(BasicClient.calledWithNew(), 'New client should be created');
 			assert.ok(BasicClient.calledWithMatch(appAuthSessionFake), 'App auth session should be passed in');
 		});
+
+		it('should throw an error when enterprise id is not passed and the SDK instance is not created from boxAppSettings', function() {
+			assert.throws(function() {
+				sdk.getAppAuthClient('enterprise');
+			}, Error, 'Enterprise ID must be passed');
+
+		});
```
mattwiller: Please also add a test that a passed-in `id` parameter is not overridden by the config value (which would have been the main impact of this bug)

KarthikShanmugasundaram: done

PR #137, opened 2017-05-30
feat: add response error message

garyforsterio:
Add response error message for clearer debugging/error handling

### DIFF
--- a/lib/box-client.js
+++ b/lib/box-client.js
@@ -643,7 +643,7 @@ BoxClient.prototype.wrapWithDefaultHandler = function(method) {
 					return response.body;
 				}
 
-				throw errors.buildResponseError(response);
+				throw errors.buildResponseError(response, response.body.code);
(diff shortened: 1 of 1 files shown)


lib/box-client.js
```diff
@@ -643,7 +643,7 @@ BoxClient.prototype.wrapWithDefaultHandler = function(method) {
 					return response.body;
 				}
 
-				throw errors.buildResponseError(response);
+				throw errors.buildResponseError(response, response.body.code);
```
mattwiller: It might actually be better to have parity with the default handler for callbacks above and do `throw errors.buildUnexpectedResponseError(response)`.  If you don't mind making that change, I'm happy to merge this after!

mattwiller:
Since this PR isn't active anymore, I implemented the requested changes in #237 and will close this PR in favor of that one.

PR #149, opened 2017-06-19
Adding recentItems route

ahsang:
adds the recent items route as per https://developer.box.com/v2.0/reference#get-recent-items
If this is fine, i can add tests and documentation for this as well

### DIFF
--- a/lib/managers/recent-items.js
+++ b/lib/managers/recent-items.js
@@ -36,6 +36,9 @@ function RecentItems(client) {
  * Method: GET
  *
  * @param {Object} [qs] - Additional options can be passed with the request via querystring. Can be left null in most cases.
+ * @param {string} [qs.limit] Maximum number of items to return
+ * @param {string} [qs.marker] The position marker for marker-based paging
+ * @param {string} [qs.fields] Comma-separated list of fields to include in the response
  * @param {Function} [callback] - Passed the items information if they were acquired successfully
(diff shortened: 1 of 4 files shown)


lib/managers/recent-items.js
```diff
@@ -0,0 +1,54 @@
...
+	this.client = client;
+}
+
+/**
+ * Requests all items that have been accessed by a user in the last 90 days or the last 1000 items accessed.
+ *
+ * API Endpoint: '/recent_items'
+ * Method: GET
+ *
+ * @param {Object} [qs] - Additional options can be passed with the request via querystring. Can be left null in most cases.
```
mattwiller: Could you please add documentation for the query string options documented at https://developer.box.com/reference#get-recent-items to the JSDoc here?  You can use the following format:

```
@param {string} [qs.marker] Marker field description goes here
```

mattwiller:
Looks good — thanks again for contributing! 🙌

ahsang:
@mattwiller can you please publish this change on npm too? i am using this for a project so would love it if the feature came in the official build.

mattwiller:
@ahsang Sure!  I'm working on #151 right now, which I would like to get released as well, so once that's finished I'll create a release with both commits in it.

mattwiller:
FYI @ahsang I just released v1.6.0 of the SDK, which includes the Recent Items endpoint.  Thanks again for contributing!

ahsang:
Thanks

PR #161, opened 2017-08-04
Fix token expiration window

mattwiller:
Fix the 150-second window between token staleness and expiration
where two Persistent Clients sharing a TokenStore could cause
one of the clients to generate 401 errors for all calls.  This
change removes the distinction between "stale" and "expired" tokens
and deprecates the staleBufferMS config option in favor of a single
expiredBufferMS option.  The old option will still be honored, but
tokens will be considered expired at the maximum of the stale
and expired times.


boxcla:

PR #162, opened 2017-08-08
Adding representations endpoint

carycheng77:


boxcla:
Verified that @carycheng77 has signed the CLA. Thanks for the pull request!

docs/files.md
```diff
@@ -746,3 +747,26 @@ A file's watermark can be removed by calling
 ```js
 client.files.removeWatermark('67890', callback);
 ```
+
+Get Representation
+------------------
+
+A file's representation can be retrieved by calling
+[`files.getRepresentations(fileID, [representationTypes], options,
```
mattwiller: Why the [] around representationTypes?

docs/files.md
```diff
@@ -746,3 +747,26 @@ A file's watermark can be removed by calling
...
+callback)`](https://opensource.box.com/box-node-sdk/Files.html#getRepresentation).
+You will be able to fetch a pdf representation, thumbnail representation, multi-page images
+representation, and fetching text representation.
+
+You can fetch different dimensions for the representation by calling
+```js
+client.files.getRepresentationInfo('67890', client.files.representation.PDF, null, callback);
```
mattwiller: This example doesn't seem to fit with the sentence above; what dimensions can they fetch here?

docs/files.md
```diff
@@ -746,3 +747,26 @@ A file's watermark can be removed by calling
...
+```
+Additionally there are two header options you can include. set_content_disposition_type and set_content_disposition_filename. set_content_disposition_type must be set to either inline
```
mattwiller: What do these options mean for Node.js?  It would be good to explain how the behavior differs for each option, since I think these are mostly for use in a browser context.

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
 };
 
 /**
+ * Enum of valid x-rep- hint values for generating representation
+ *
+ * @readonly
+ * @enum {FileRepresentationType}
```
mattwiller: Where is this type defined?  You may need to put a `@typedef` somewhere to describe what this type is

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+ * @param {string} [options.set_content_disposition_type] - Set to either 'inline' for the file to be downloaded from the
+ *    browser or 'attachment' which opens the file in another tab
+ * @param {string} [options.set_content_disposition_filename] - If defined this will name the downloaded representation file, if not the name will
+ *    be the original Box file name
+ * @param {string} [options.asset_path] - This optional field is for requesting a multi-page representation. This value should be in format of
+			<page_number>.<png>. 4-page long document will be named: 1.png, 2.png, etc...
+ * @param {Function} [callback] - Passed  a url link to download the representation that was generated
+ * @returns {Promise<Object>} A promise resolving to the representation response objects
+*/
+Files.prototype.getRepresentationInfo = function(fileID, representationType, options, callback) {
```
mattwiller: This method name doesn't match the one in the docs. Which one did you want to use?

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+ * @param {FileRepresentationType} representationType - The x-rep-hints value the application should create a
+ *    representation for. This value should be access from enum
+ * @param {Object} [options] - Additional options. Can be left null in most cases.
+ * @param {string} [options.set_content_disposition_type] - Set to either 'inline' for the file to be downloaded from the
+ *    browser or 'attachment' which opens the file in another tab
+ * @param {string} [options.set_content_disposition_filename] - If defined this will name the downloaded representation file, if not the name will
+ *    be the original Box file name
+ * @param {string} [options.asset_path] - This optional field is for requesting a multi-page representation. This value should be in format of
+			<page_number>.<png>. 4-page long document will be named: 1.png, 2.png, etc...
+ * @param {Function} [callback] - Passed  a url link to download the representation that was generated
```
mattwiller: This param description doesn't match what's actually passed to the callback; it should receive the whole representations object, not just the URL string

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+};
+/**
+ * Requests links to all representations generated for a specific Box file
+ *
+ * API Endpoint: '/files/:fileID?fields=representations'
+ * Method : GET
+ *
+ * @param {string} fileID - Box ID of the file being requested
+ * @param {FileRepresentationType} representationType - The x-rep-hints value the application should create a
+ *    representation for. This value should be access from enum
```
mattwiller: They shouldn't *need* to use the enum; a user could always provide their own value if they want

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+ * @param {Object} [options] - Additional options. Can be left null in most cases.
+ * @param {string} [options.set_content_disposition_type] - Set to either 'inline' for the file to be downloaded from the
```
mattwiller: Since this code probably isn't running in a browser, we should be really clear about what this option does in Node

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+ * @param {string} [options.set_content_disposition_filename] - If defined this will name the downloaded representation file, if not the name will
+ *    be the original Box file name
+ * @param {string} [options.asset_path] - This optional field is for requesting a multi-page representation. This value should be in format of
+			<page_number>.<png>. 4-page long document will be named: 1.png, 2.png, etc...
+ * @param {Function} [callback] - Passed  a url link to download the representation that was generated
+ * @returns {Promise<Object>} A promise resolving to the representation response objects
+*/
+Files.prototype.getRepresentationInfo = function(fileID, representationType, options, callback) {
+	var assetPath = options.asset_path || '';
+	var template_url = '';
```
mattwiller: You don't really need to define this here

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+	var params = {
+		qs: {
+			fields: 'representations'
+		},
+		headers: {
+			'x-rep-hints': representationType
+		}
+	};
+	delete options.asset_path;
```
mattwiller: Maybe move this `delete` statement to under where you define `assetPath` for clarity, so it's obvious you're just moving the value off to a local variable

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+	Object.assign(params.headers, options);
+	var apiPath = urlPath(BASE_PATH, fileID);
+
+	return this.client.get(apiPath, params)
+	.then(response => {
```
mattwiller: [style] We usually indent chained methods one level when they're on a different line

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+		switch (response.statusCode) {
+			// 202 - A Box file representation will be generated, but is not ready yet
+		case httpStatusCodes.ACCEPTED:
```
mattwiller: [style] The `case` statements should probably be indented so the line up with the comments on the lines above

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+			throw errors.buildResponseError(response, 'Representation not ready at this time');
+
+			// 200 - A Boxfile representation image received
+			// return the representation file
+		case httpStatusCodes.OK:
+			response.body.representations.entries.forEach(function(representation) {
+				template_url = representation.content.url_template;
+				template_url.replace('{+asset_path}', assetPath);
```
mattwiller: I don't think this does what you expect it to; `String#replace()` returns a new string with the replaced value and doesn't modify the original string (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)

lib/managers/files.js
```diff
@@ -1243,6 +1243,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+		switch (response.statusCode) {
+			// 202 - A Box file representation will be generated, but is not ready yet
+		case httpStatusCodes.ACCEPTED:
+			throw errors.buildResponseError(response, 'Representation not ready at this time');
+
+			// 200 - A Boxfile representation image received
+			// return the representation file
+		case httpStatusCodes.OK:
+			response.body.representations.entries.forEach(function(representation) {
+				template_url = representation.content.url_template;
```
mattwiller: There's no need to have this variable defined in the upper scope; you could just use `var` or `let` here

tests/lib/managers/files-test.js
```diff
@@ -2440,7 +2440,7 @@ describe('Files', function() {
 
 		it('should retry the call when the API returns a 202 with Retry-After header', function(done) {
 
-            // Need to fake timers and make Promises resolve synchronously for this test to work
+						// Need to fake timers and make Promises resolve synchronously for this test to work
```
mattwiller: The indentation on this line looks like it got messed up

tests/lib/managers/files-test.js
```diff
@@ -2952,4 +2952,144 @@ describe('Files', function() {
 
 	});
 
+	describe('getRepresentationInfo()', function() {
+		var testRepresentation = '[png?dimensions=1024x1024]';
```
mattwiller: [style] For constant values like this that don't change between tests, we usually use UPPER_CASE variable names

tests/lib/managers/files-test.js
```diff
@@ -2952,4 +2952,144 @@ describe('Files', function() {
...
+								}
+							}
+						]
+					}
+				}
+			};
+			sandbox.stub(boxClientFake, 'get').returns(Promise.resolve(response));
+			files.getRepresentationInfo(FILE_ID, testRepresentation, representationOptions, function(err, representationObject) {
+				assert.ifError(err);
+				assert.strictEqual(representationObject, response.body.representations, 'representation object is returned');
```
mattwiller: This tests that the objects are the same (which is good), but it doesn't do any assertion about the asset path replacement.  You should add another assertion to verify that the new `asset_url` property that was added to the object is what you expect.

tests/lib/managers/files-test.js
```diff
@@ -2952,4 +2952,144 @@ describe('Files', function() {
...
+								}
+							}
+						]
+					}
+				}
+			};
+			sandbox.stub(boxClientFake, 'get').returns(Promise.resolve(response));
+			return files.getRepresentationInfo(FILE_ID, testRepresentation, representationOptions)
+				.then(representationObject => {
+					assert.strictEqual(representationObject, response.body.representations, 'representation object is returned');
```
mattwiller: Same as above — there should also be an assertion about the `asset_url` property here

tests/lib/managers/files-test.js
```diff
@@ -2952,4 +2952,144 @@ describe('Files', function() {
...
+			var apiError = new Error('ECONNRESET');
+			sandbox.stub(boxClientFake, 'get').returns(Promise.reject(apiError));
+
+			return files.getRepresentationInfo(FILE_ID, testRepresentation, representationOptions)
+				.catch(err => {
+					assert.equal(err, apiError);
+				});
+		});
+
+		it('should call callback with unexpected response error when the API returns unknown status code', function() {
```
mattwiller: This test case is missing a Promise version

lib/managers/files.js
carycheng77: Hey Matt! by docs did you mean the files.md or the developer docs

docs/files.md
```diff
@@ -746,3 +747,35 @@ A file's watermark can be removed by calling
...
+representation, and fetching text representation.
+
+You can fetch different dimensions for the representation by calling
+```js
+client.files.getRepresentationInfo('67890', client.files.representation.IMAGE_LARGE, null, callback);
+```
+Additionally there are two header options you can include. set_content_disposition_type and set_content_disposition_filename.
+
+set_content_disposition_type must be set to either 'inline' to try
+and open the representation within the browser or 'attachment' to force the browser to download the
```
mattwiller: I still think we need to get more clarity about exactly what these parameters mean for someone using the Node SDK.  If you call with one vs. the other, what is the actual difference in behavior?  I think talking about the browser here might confuse users who aren't in a browser.

I'm also really confused about what these mean for this specific endpoint.  This is not the call for the actual asset, it's for the JSON object describing the representations.  Do these headers even apply here, or are they for the actual asset request itself?

docs/files.md
```diff
@@ -746,3 +747,35 @@ A file's watermark can be removed by calling
...
+representation. If not supplied the user-agent of the client is allowed to interpret that however
+it wants. To ensure consistency, we recommend supplying this parameter.
```
mattwiller: If this is important to provide for consistency, we may want to set a sane default for this SDK.

docs/files.md
```diff
@@ -746,3 +747,35 @@ A file's watermark can be removed by calling
...
+
+set_content_disposition_filename allows the application to define the downloaded representation's
```
mattwiller: Does this option only make sense if the user has already specified `set_content_disposition: attachment`?  What does this do if they specified `inline`?

lib/managers/files.js
```diff
@@ -3,6 +3,14 @@
  */
 
 'use strict';
+// -----------------------------------------------------------------------------
+// Typedefs
+// -----------------------------------------------------------------------------
+
+/**
+ * A webhook trigger type constant
+ * @typedef {FileRepresentationType} Different representations we can request from reps endpoint
```
mattwiller: The correct format here is `@typedef {string} FileRepresentationType` — see http://usejsdoc.org/tags-typedef.html

lib/managers/files.js
```diff
@@ -1243,6 +1251,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+	EXTRACTED_TEXT: '[extracted_text]'
+};
+/**
+ * Requests links to all representations generated for a specific Box file
+ *
+ * API Endpoint: '/files/:fileID?fields=representations'
+ * Method : GET
+ *
+ * @param {string} fileID - Box ID of the file being requested
+ * @param {string} representationType - The x-rep-hints value the application should create a
```
mattwiller: Shouldn't this be `@param {FileRepresentationType}`?

lib/managers/files.js
```diff
@@ -1243,6 +1251,78 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+		}
+	};
+	Object.assign(params.headers, options);
+	var apiPath = urlPath(BASE_PATH, fileID);
+
+	return this.client.get(apiPath, params)
+		.then(response => {
+			switch (response.statusCode) {
+				// 202 - A Box file representation will be generated, but is not ready yet
+			case httpStatusCodes.ACCEPTED:
```
mattwiller: [style] These should still be indented to line up with the comments on the line above

tests/lib/managers/files-test.js
```diff
@@ -2952,4 +2952,164 @@ describe('Files', function() {
...
+		it('should call callback with the representation when a 200 response is returned', function(done) {
+			var response = {
+				statusCode: 200,
+				body: {
+					representations: {
+						entries: [
+							{
+								content: {
+									url_template: placeholderUrl,
+									asset_url: placeholderUrlWithoutAssetPath
```
mattwiller: This isn't quite right; this value isn't returned in the API response.  It's best to make the test cases line up with how the API actually responds to avoid introducing weird bugs.

tests/lib/managers/files-test.js
```diff
@@ -2952,4 +2952,164 @@ describe('Files', function() {
...
+		it('should return a promise resolving to a representation object when a 200 response is returned', function() {
+			var response = {
+				statusCode: 200,
+				body: {
+					representations: {
+						entries: [
+							{
+								content: {
+									url_template: placeholderUrl,
+									asset_url: placeholderUrlWithoutAssetPath
```
mattwiller: Ditto here — this isn't in the API response

docs/files.md
```diff
@@ -746,3 +747,24 @@ A file's watermark can be removed by calling
...
 ```
+
+Get Representation Info
+-----------------------
+
+A file's representation info can be retrieved by calling
+[`files.getRepresentationInfo(fileID, representationTypes
+callback)`](https://opensource.box.com/box-node-sdk/Files.html#getRepresentationInfo).
+You will be able to fetch information regarding pdf representation, thumbnail representation, multi-page images
+representation, and extracting text representation.
```
mattwiller: [typo] You probably want "extracted text" here

docs/files.md
```diff
@@ -746,3 +747,24 @@ A file's watermark can be removed by calling
...
+
+You can retrieve information regarding the generated representations by calling. This will retrieve information
+for a 2048x2048 jpg representation and a 2048x2038 png representation generated for your Box file.
```
mattwiller: [typo] You have 2038 instead of 2048 here

docs/files.md
```diff
@@ -37,6 +37,7 @@ file's contents, upload new versions, and perform other common file operations
 * [Get Watermark](#get-watermark)
 * [Apply Watermark](#apply-watermark)
 * [Remove Watermark](#remove-watermark)
+* [Get Representation](#get-representation)
```
mattwiller: This needs to be updated to match the heading below

carycheng77:
@mattwiller Thanks Matt! I just fixed those, appreciate you looking these over!

lib/managers/files.js
```diff
@@ -3,6 +3,14 @@
  */
 
 'use strict';
+// -----------------------------------------------------------------------------
+// Typedefs
+// -----------------------------------------------------------------------------
+
+/**
+ * A webhook trigger type constant
```
mattwiller: This comment needs to be updated

docs/files.md
```diff
@@ -746,3 +747,24 @@ A file's watermark can be removed by calling
...
+You will be able to fetch information regarding pdf representation, thumbnail representation, multi-page images
+representation, and extracted text representation.
+
+You can retrieve information regarding the generated representations by calling. This will retrieve information
+for a 2048x2048 jpg representation and a 2048x2048 png representation generated for your Box file.
+```js
+client.files.getRepresentationInfo('67890', client.files.representation.IMAGE_LARGE, callback);
+```
+
+Similarly you can form your own request by manually passing in the representation types you want to
```
mattwiller: It might be good to link to the API documentation (https://developer.box.com/reference#section-x-rep-hints-header) here so users know what the format of this should look like.

lib/managers/files.js
```diff
@@ -1243,6 +1251,61 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+};
+/**
+ * Requests information for all representation objects generated for a specific Box file
+ *
+ * API Endpoint: '/files/:fileID?fields=representations'
+ * Method : GET
+ *
+ * @param {string} fileID - Box ID of the file being requested
+ * @param {FileRepresentationType} representationType - The x-rep-hints value the application should create a
+ *    representation for. This value can either come from FileRepresentationType enum or manually created
```
mattwiller: It might be good to actually show where the enum is (i.e. `client.files.representation`)

lib/managers/files.js
```diff
@@ -1243,6 +1251,61 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+ * @param {Function} [callback] - A representation object containing the url_template, info, properties, and status
```
mattwiller: This description is kind of weird; it's describing the value passed to the callback not the callback itself.  Also, the value passed to the callback isn't a single representation, it's a collection of representation objects

lib/managers/files.js
```diff
@@ -1243,6 +1251,61 @@ Files.prototype.getCollaborations = function(fileID, options, callback) {
...
+		.then(response => {
+			switch (response.statusCode) {
+			// 202 - A Box file representation will be generated, but is not ready yet
+			case httpStatusCodes.ACCEPTED:
+				throw errors.buildResponseError(response, 'Representation not ready at this time');
+
+			// 200 - A Boxfile representation generated
+			// return the representation object
+			case httpStatusCodes.OK:
+				return response.body.representations;
```
mattwiller: I just remembered that this return value is a collection it probably needs a case to handle if the user has the `iterators` option turned on!  See the [default response handler](https://github.com/box/box-node-sdk/blob/master/lib/box-client.js#L661) for an example of what we're doing, we need to do something similar here since.

carycheng77: @mattwiller I took a iterators check from box-clients, let me know if that looks right!

PR #163, opened 2017-08-09
Implement Batch API

mattwiller:
Adding support for batching API calls via a mode on the client,
e.g. client.batch()  Each call still provides its own callback/then
to prevent needing to index into an array of responses, but the
entire batch response is provided after the batch is executed via
client.execute() in case users want to consume the results in
aggregate.


boxcla:

docs/client.md
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    console.log(response.body);
+});
+```
+
+### DELETE
+
+To make a DELETE call, call the
+[`client.del(url, params, callback)`](http://opensource.box.com/box-node-sdk/BoxClient.html#del)
```
mattwiller: Back when I was a noob I didn't think you could use `delete` as a property name :(

PR #165, opened 2017-08-10
Control number of API calls made by Event Stream

mattwiller:
Refactor the Event Stream class to use Promises internally, and
make two substantive changes to help control the number of API
calls made behind the scenes and help prevent hitting rate limits:

1. Increase the limit value of the events.get() call from the
default of 100 to 500
2. Add a rate limiter to ensure that events are fetched at most
once per second


boxcla:

lib/event-stream.js
```diff
@@ -138,84 +145,94 @@ EventStream.prototype.doLongPoll = function() {
...
-		}
-
-		// Push new events into the stream
-		newEvents.forEach(function(event) {
-			self._dedupHash[event.event_id] = true;
-			self.push(event);
-		});
+	var eventParams = {
+		stream_position: this._streamPosition,
+		limit: 500
```
djordan: Should this limit be configurable?

mattwiller: I don't think so; for performance we should always take the maximum.  We do the same in the Admin Events stream as well.

PR #174, opened 2017-10-04
Allow for multiple scopes tokenexchange (fixes #173)

### LINKED ISSUE #173: exchangeToken method does not work for multiple scopes
In token-manager.js, the exchangeToken manager does work with multiple scopes.

The reason is because of this line of code:
`scope: (typeof scopes === 'string' ? scopes : scopes.join(','))`

Needs to be changed to:
`scope: (typeof scopes === 'string' ? scopes : scopes.join(' '))`

scope parameter needs to be in space separated rather than comma separated for exchange.

I have written a fix. I am about to push fix from my forked SDK.

cghil:
Here is an example curl command for a working token exchange.

```
curl -X POST \
  https://api.box.com/oauth2/token \
  -F subject_token=<ACCESS_TOKEN> \
  -F subject_token_type=urn:ietf:params:oauth:token-type:access_token \
  -F 'scope=base_preview annotation_view_all annotation_edit' \
  -F grant_type=urn:ietf:params:oauth:grant-type:token-exchange \
  -F resource=https://api.box.com/2.0/files/<FILE_ID>
```

See how scopes use spaces, not commas.

### DIFF
--- a/lib/token-manager.js
+++ b/lib/token-manager.js
@@ -362,7 +362,7 @@ TokenManager.prototype = {
 			grant_type: grantTypes.TOKEN_EXCHANGE,
 			subject_token_type: ACCESS_TOKEN_TYPE,
 			subject_token: accessToken,
-			scope: (typeof scopes === 'string' ? scopes : scopes.join(','))
+			scope: (typeof scopes === 'string' ? scopes : scopes.join(' '))
 		};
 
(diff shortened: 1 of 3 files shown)


boxcla:
Verified that @cghil has signed the CLA. Thanks for the pull request!

Issue #177, opened 2017-10-13
getAllFileVersionRetentions - Unable to determine paging strategy for response


kylefernandadams:
getAllFileVersionRetentions is unable to determine paging strategy:
```
Unable to determine paging strategy for response!
    at new PagingIterator (/Developer/Code/box-governance-center/server/node_modules/box-node-sdk/lib/util/paging-iterator.js:84:10)
    at ret.then.response (/Developer/Code/box-governance-center/server/node_modules/box-node-sdk/lib/box-client.js:654:14)
    at tryCatcher (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:693:18)
    at Promise._fulfill (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:638:18)
    at Promise._settlePromise (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:582:21)
    at Promise._settlePromise0 (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:693:18)
    at Promise._fulfill (/Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/promise.js:638:18)
    at /Developer/Code/box-governance-center/server/node_modules/bluebird/js/release/nodeback.js:42:21
    at BoxClient._handleResponse (/Developer/Code/box-governance-center/server/node_modules/box-node-sdk/lib/box-client.js:308:2)
    at handleResponseCallback (/Developer/Code/box-governance-center/server/node_modules/box-node-sdk/lib/box-client.js:241:10)
    at /Developer/Code/box-governance-center/server/node_modules/box-node-sdk/lib/api-request.js:285:3
```

mattwiller:
Thanks for reporting — I'll take a look!

[...]

PR #179, opened 2017-10-17
Fix paging iterator for missing next_marker cases

mattwiller:
Some Box API endpoints do not return any `next_marker` field when the collection has finished paging; in this case, we should treat the collection as finished, buffer the data in the response, and let the user iterate over the buffer.

### DIFF
--- a/lib/util/paging-iterator.js
+++ b/lib/util/paging-iterator.js
@@ -72,7 +72,6 @@ class PagingIterator {
 			throw new Error('Cannot create paging iterator for non-paged response!');
 		}
 
-		this.done = false;
 
 		var data = response.body;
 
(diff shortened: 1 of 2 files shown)


boxcla:

mattwiller:
Fixes #177

lib/util/paging-iterator.js
```diff
@@ -81,11 +83,13 @@ class PagingIterator {
 			this.nextField = PAGING_MODES.MARKER;
 			this.nextValue = data.next_marker;
 		} else {
-			throw new Error('Unable to determine paging strategy for response!');
+			// throw new Error('Unable to determine paging strategy for response!');
+			this.nextField = PAGING_MODES.MARKER;
+			this.nextValue = null;
+			this.done = true;
```
jmoldow: Why do you need this here?

Wouldn't it get set to `true` in the same way as if `data.next_marker` were `null` in the `else if` block?

mattwiller:
@jmoldow You're right!  I removed the unnecessary `done` setting and added a comment to clarify what this case is doing.

PR #182, opened 2017-11-01
Add newer Node.js versions to CI

mattwiller:


boxcla:

mattwiller:
@djordan I thought about that, but since those versions were originally supported I think it would be a breaking change if they no longer worked.  We might consider removing support for them in a major version bump sometime in the future.

PR #183, opened 2017-11-03
init for terms of service

carycheng77:
Thanks Matt!


boxcla:
Verified that @carycheng77 has signed the CLA. Thanks for the pull request!

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,251 @@
...
+		};
+		Object.assign(params.body, user);
+	}
+
+	var apiPath = urlPath(USER_STATUSES_PATH);
+	return this.client.wrapWithDefaultHandler(this.client.post)(apiPath, params, callback);
+};
+
+/**
+ * Gets custom terms of services given terms of services id and user id
```
mattwiller: This description doesn't look accurate — this method doesn't take a user ID

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,500 @@
...
+					text: 'TEST',
+					created_at: '2017-10-21T20:09:59-07:00',
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.stub(boxClientFake, 'get').returns(Promise.resolve(response));
+			return termsOfService.getTermsOfServicesByID(TERMS_OF_SERVICE_ID)
+				.then(termsOfServiceObject => {
+					assert.strictEqual(termsOfServiceObject, response, 'terms of service object is returned');
```
mattwiller: This assertion doesn't look like it's verifying the correct behavior, since the expected value should be just the body instead of the full response.

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,500 @@
...
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.stub(boxClientFake, 'get').returns(Promise.resolve(response));
+			termsOfService.getTermsOfServicesByID(TERMS_OF_SERVICE_ID, function(err, termsOfServiceObject) {
+				assert.ifError(err);
+				assert.strictEqual(termsOfServiceObject, response.body, 'terms of service object is returned');
+			});
+			done();
```
mattwiller: The `done()` call shouldn't go here — it needs to go inside the callback to make the test wait until after the Promise resolves

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,500 @@
...
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.stub(boxClientFake, 'get').returns(Promise.resolve(response));
+			termsOfService.getTermsOfServices(function(err, termsOfServiceObject) {
+				assert.ifError(err);
+				assert.strictEqual(termsOfServiceObject, response.body, 'terms of service object is returned');
+			});
+			done();
```
mattwiller: It looks like many of these tests have some of the same issues as the first set above; please update them all and let me know if you run into difficulties!

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,500 @@
...
+		});
+	});
+
+
+	describe('getTermsOfServices()', function() {
+		it('should make GET request to get terms of service info when called', function() {
+			var response = {
+				statusCode: 200,
+				body: {
+					total_count: 2,
```
mattwiller: I don't think it matters, but this `total_count` doesn't match the array below, which might be confusing for someone coming into this file later

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,500 @@
...
+					status: 'enabled',
+					enterprise: {},
+					tos_type: 'managed',
+					text: 'This is a test',
+					created_at: '2017-10-21T20:09:59-07:00',
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.mock(boxClientFake).expects('post').withArgs('/terms_of_services')
```
mattwiller: This mock doesn't assert anything about the body parameters — those should be added to `withArgs()` to ensure that the correct parameters are being constructed and passed to the API.  This should also be done for all the other endpoints that take body params

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,251 @@
...
+	};
+
+	var params = {
+		body: {
+			tos: tos,
+			is_accepted: isAccepted
+		}
+	};
+
+	if (options) {
```
mattwiller: This condition should probably be `if (options.userID)` in case the user passes an empty object, which may happen for callback users

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,251 @@
...
+		var user = {
+			user_id: options.termsOfServiceUserID,
+			type: 'user'
+		};
+		Object.assign(params.body, user);
```
mattwiller: I don't think this does the right thing; the user object should probably be attached as a single property

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * Accepts/rejects custom terms of services for the user
+ *
+ * API Endpoint: '/terms_of_service_user_statuses'
+ * Method: POST
+ *
+ * @param {string} termsOfServicesID - Terms of services ID to retrieve user statuses on
+ * @param {bool} isAccepted - Determines wehether the terms of services has been accepted or rejected
+ * @param {Object} [options] - Additional options. Can be left null in most cases.
+ * @param {string} [options.termsOfServiceUserID] - Optional, user id to retrieve terms of service status on, default is current user
```
mattwiller: Consider renaming this parameter to something simpler, e.g. `userID` for consistency with the method below

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,251 @@
...
+ * @param {Function} [callback] - Passed the terms of service user status information if successful, error otherwise
+ * @returns {Promise<Object>} A promise resolving to the terms of service user status
+ */
+TermsOfService.prototype.createTermsOfServiceUserStatuses = function(termsOfServicesID, isAccepted, options, callback) {
+	var tos = {
```
mattwiller: Consider inlining this object definition into the body below so it's easier to see the full shape of the payload

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,251 @@
...
+		tos_id: termsOfServicesID,
+		type: 'terms of service'
```
mattwiller: Is this the expected `type` value?  I've never seen one with spaces before

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,251 @@
...
+};
+
+/**
+ * Gets custom terms of services for the user's enterprise
+ *
+ * API Endpoint: '/terms_of_services'
+ * Method: GET
+ *
+ * @param {Object} [options] - Additional options. Can be left null in most cases.
+ * @param {string} [options.tosType] - Optional, indicates whether the terms of service is set for external or managed under enterprise
```
mattwiller: This should use the `{TermsOfServicesType}` type

carycheng77:
Hey @mattwiller could you take a look at line 276 in terms-of-service manager? I don't think I am assigning Object correctly here, thanks!

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,307 @@
...
+			is_accepted: isAccepted
+		}
+	};
+
+	if (options.userID) {
+		var user = {
+			user_id: options.userID,
+			type: 'user'
+		};
+		Object.assign(params.body.user, user);
```
mattwiller: This should probably be `params.body.user = user`

lib/managers/terms-of-service.js
```diff
@@ -0,0 +1,307 @@
...
+			is_accepted: isAccepted
+		}
+	};
+
+	if (options.userID) {
+		var user = {
+			user_id: options.userID,
+			type: 'user'
+		};
+		params.body.user = user;
```
mattwiller: [nit] You don't need to create a `user` variable here, and could assign the object directly to `params.body.user` instead

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+				enterprise: {},
+				tos_type: 'external',
+				text: 'TEST',
+				created_at: '2017-10-21T20:09:59-07:00',
+				modified_at: '2017-10-21T20:09:59-07:00'
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.mock(boxClientFake).expects('get').withArgs('/terms_of_services/1234')
+				.returns(Promise.resolve(response));
+			termsOfService.get(TERMS_OF_SERVICE_ID);
```
mattwiller: For completeness, we should pass options here and ensure they get passed through to `get()`

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+					tos_type: 'external',
+					text: 'TEST',
+					created_at: '2017-10-21T20:09:59-07:00',
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}]
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.mock(boxClientFake).expects('get').withArgs('/terms_of_services')
+				.returns(Promise.resolve(response));
+			termsOfService.getAll();
```
mattwiller: For completeness, we should pass options here and make sure they get passed through to `get()`

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+			sandbox.stub(boxClientFake, 'get').returns(Promise.resolve(response));
+			return termsOfService.getAll()
+				.then(termsOfServiceObject => {
+					assert.strictEqual(termsOfServiceObject, response, 'terms of service object is returned');
+				});
+		});
+	});
+
+	describe('create()', function() {
+		var expectedTermsOfServiceParam = { body: { status: 'enabled', text: 'Test Text', tos_type: 'managed'}};
```
mattwiller: It looks like you're just using this in the first test, so you can pull it into that `it` block

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.stub(boxClientFake, 'put').returns(Promise.resolve(response));
+			return termsOfService.update(TERMS_OF_SERVICE_ID, 'enabled', 'Test Text')
+				.then(termsOfServiceObject => {
+					assert.strictEqual(termsOfServiceObject, response, 'terms of service object is returned');
+				});
+		});
+	});
+
+	describe('createUserStatuses()', function() {
```
mattwiller: When you change the name of this method to make it singular, remember to change it here too!

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+				tos: {},
+				user: {},
+				is_accepted: true,
+				created_at: '2017-10-21T20:09:59-07:00',
+				modified_at: '2017-10-21T20:09:59-07:00'
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.mock(boxClientFake).expects('post').withArgs('/terms_of_service_user_statuses', expectedParam)
+				.returns(Promise.resolve(response));
+			termsOfService.createUserStatuses(TERMS_OF_SERVICE_ID, true, TERMS_OF_SERVICE_USER_ID);
```
mattwiller: You're passing in a user ID here, but I don't see that reflected in the expected body params.  Make sure that this is being correctly passed through!

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+	describe('getUserStatuses()', function() {
+		it('should make GET request to retrieve custom terms of service user status when called', function() {
+			var response = {
+				total_count: 2,
+				entries: []
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.mock(boxClientFake).expects('get').withArgs('/terms_of_service_user_statuses')
+				.returns(Promise.resolve(response));
+			termsOfService.getUserStatuses(TERMS_OF_SERVICE_ID);
```
mattwiller: You're passing an ID here but not verifying it anywhere else in the test — need to make sure it's being correctly used

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+					user: {},
+					is_accepted: true,
+					created_at: '2017-10-21T20:09:59-07:00',
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.mock(boxClientFake).expects('post').withArgs('/terms_of_service_user_statuses', expectedParam)
+				.returns(Promise.resolve(response));
+			termsOfService.setUserStatuses(TERMS_OF_SERVICE_ID, true, TERMS_OF_SERVICE_USER_ID);
```
mattwiller: You're setting a user ID here but not verifying it anywhere in the test — need to make sure it's being correctly used

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+		});
+
```
mattwiller: For these tests with optional params, you'll probably need to create multiple tests to achieve full code coverage: one with the parameter passed, and one without

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+					type: 'terms_of_service_user_status',
+					id: '1234',
+					tos: {},
+					user: {},
+					is_accepted: true,
+					created_at: '2017-10-21T20:09:59-07:00',
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
```
mattwiller: `wrapWithDefaultHandler` isn't being called in this tests, so you should remove the stub for it in all of them

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+			sandbox.stub(boxClientFake, 'post').returns(Promise.resolve(response));
+			return termsOfService.setUserStatuses(TERMS_OF_SERVICE_ID, true, TERMS_OF_SERVICE_USER_ID)
+				.then(termsOfServiceUserStatusObject => {
+					assert.strictEqual(termsOfServiceUserStatusObject, response.body, 'terms of service user status object is returned');
+				});
+		});
+
+		it('should make PUT request to retrieve custom terms of service user status when called', function() {
```
mattwiller: This test name should be updated to include the circumstances when the PUT call would be made, e.g. "should make PUT request to update user status when user status already exists"

tests/lib/managers/terms-of-service-test.js
```diff
@@ -0,0 +1,571 @@
...
+					id: '1234',
+					tos: {},
+					user: {},
+					is_accepted: true,
+					created_at: '2017-10-21T20:09:59-07:00',
+					modified_at: '2017-10-21T20:09:59-07:00'
+				}
+			};
+			sandbox.stub(boxClientFake, 'wrapWithDefaultHandler').returnsArg(0);
+			sandbox.mock(boxClientFake).expects('put').withArgs('/terms_of_service_user_statuses/1234')
```
mattwiller: How is this test passing?  You're not stubbing POST anywhere...

PR #191, opened 2017-12-07
Add optional parameters to file upload methods

### LINKED ISSUE #190: uploadFile does not support content_created_at or content_modified_at
In the REST API docs, https://developer.box.com/v2.0/reference#upload-a-file documents the `content_created_at` and `content_modified_at` fields. However, there is nowhere to specify these fields in the uploadFile() function in the Box Node SDK. Indeed, there is a comment in the code stating "This method currently does not support any optional parameters such as contentModifiedAt"

For example, the uploadFile function could take an object with additional properties to be included in the request, similar to the `qs` parameter in the `files.get` function. Of course, it would still have to work if the object parameter were not provided.

mattwiller:
Fixes #190

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -147,6 +147,17 @@ var buffer = new Buffer(50);
 client.files.uploadFile('98768', 'New File', buffer, callback);
 ```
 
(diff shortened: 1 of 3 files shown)


boxcla:

PR #206, opened 2018-01-10
support for metadata template deletion

carycheng:


boxcla:
Hi @carycheng, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

docs/metadata.md
```diff
@@ -113,3 +114,14 @@ method.
...
+
+Delete Metadate Template
+------------------------
+
+To delete a metadata template call the
+[`metadata.deleteTemplate(scope, template, callback)`](http://opensoure.box.com/box-node-sdk/Metadata.html#deleteTemplate)
+method with the template scope and template name.
+
+```js
+client.metadata.deleteMetadata('enterprise', 'testtemplate', callback);
```
mattwiller: This example needs to be updated to use the correct method name from above — should be `deleteTemplate()`

lib/managers/metadata.js
```diff
@@ -165,6 +165,24 @@ Metadata.prototype = {
...
+	/**
+	 * Delete a metadata template from an enterprise.
+	 *
+	 * API Endpoint: '/metadata_templates/:scope/:template/schema'
+	 * Method: DELETE
+	 *
+	 * @param {string} scope - The scope of the template to delete
+	 * @param {string} template - The template to delete
+	 * @param {Function} [callback] - Passed empty response body if successful, err otherwise
+	 * @returns {Promise<Object>} A promise resolving to the updated template
```
mattwiller: This actually returns `{Promise<void>}` since there's no value coming back from the API

tests/lib/managers/metadata-test.js
```diff
@@ -214,6 +214,52 @@ describe('Metadata', function() {
...
 	});
 
+	describe('deleteTemplate()', function() {
+		var scope,
+			template;
+
+		beforeEach(function() {
+
+			scope = 'enterprise';
+			template = 'testtemplate';
```
mattwiller: If these are constant, it's okay to just declare them statically on lines 217 and 218 above, e.g.
```js
var scope = 'enterprise',
	template = 'testtemplate';
```

PR #210, opened 2018-01-20
Fix chunked uploader on streams from request library

### LINKED ISSUE #207: Issue using chunked uploader
I'm trying to use the chunked upload API (https://developer.box.com/reference#chunked-upload) via the Box Node SDK (https://github.com/box/box-node-sdk/blob/master/docs/files.md#chunked-upload) to upload large files. I'm using request (https://www.npmjs.com/package/request#streaming) to get an http.IncomingMessage (https://nodejs.org/api/http.html#http_class_http_incomingmessage), which implements the ReadableStream interface, then passing it as an argument to the Box Node client, like so:

```
request.get(url).on('response', readableStream => { 
    this._boxClient.getChunkedUploaderAsync('0', filesize, filename, readableStream, null, (err, uploader) => { 
        if (err) {...}

        uploader.on('error', err => {...}); 
        uploader.on('uploadComplete', file => {...});

        uploader.on('uploaded', data => { 
            h.console.logRed('data of the uploaded chunk', data); 
        }); 
        uploader.on('chunkUploaded', data => { 
            h.console.logRed('data of the uploaded chunk', data); 
        }); 
        uploader.on('chunkError', err => { 
            h.console.logRed('chunkError', err); 
        });

        uploader.start(); 
    }); 
});
```

However, when I try to upload a ~50mb PDF, I find that the first 5 chunks upload successfully, but the last fails with a 400 (at least judging by the 'offset' field). I couldn't find a helpful error message, but here's the logging from the above code: https://www.dropbox.com/s/s1pgzm59wmeuykl/log.txt?dl=0

Here is the file I was using to test: https://www.dropbox.com/s/w7kxrdv31tliblb/Artificial%20Intelligence%20A%20Modern%20Approach%203e%20%282009%29.pdf?dl=0

mattwiller:
It looks like request places the stream into flowing mode by
default sometime after the "response" event is fired. This causes
inconsistency in trying to read the stream manually later after the
upload session is created since some bytes may have already been
read from the stream. This appears to only be an issue with streams
from request; streams from e.g. fs.createReadStream() or manually
using the Node HTTP library to get an IncomingMessage directly
don't have this problem because nothing places them into flowing
mode. I suspect this is due to request providing both a stream
interface and a callback interface via the same function call — in
order to pass the full response to a callback it would need to
start reading data from the underlying IncomingMessage at some
point.  This fix pauses the stream BEFORE creating the upload
session to avoid this problem.

Fixes #207

### DIFF
--- a/lib/managers/files.js
+++ b/lib/managers/files.js
@@ -32,6 +32,7 @@ var urlPath = require('../util/url-path'),
 	httpStatusCodes = require('http-status'),
 	crypto = require('crypto'),
 	Promise = require('bluebird'),
+	Readable = require('stream').Readable,
 	ChunkedUploader = require('../chunked-uploader');
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


boxcla:

PR #211, opened 2018-01-23
Use more informative param names in docs and code

mattwiller:
Switch all uses of `qs` and `queryString` to options to be more
generic (since the fact that these options are passed in the query
string is an implementation detail).  Also changes `options` in
update() methods to `updates` for clarity.


boxcla:

PR #216, opened 2018-01-27
Added helper function to get box authorization token of client instance

lyaa:
This functionality is useful for apps that build on the Box Platform.

### DIFF
--- a/lib/box-client.js
+++ b/lib/box-client.js
@@ -503,6 +503,16 @@ BoxClient.prototype.exchangeToken = function(scopes, resource, options, callback
 };
 
 /**
+ * Returns a valid box token corresponding to the box client instance
+ *
+ * @returns {Promise} A promise resolving to a token for the client instance
+ */
+BoxClient.prototype.getToken = function(){
+	return this._session.getAccessToken(this._tokenOptions)
+}
+
+
+/**
  * Makes GET request to Box API V2 endpoint
  *
  * @param {string} path - path to a certain API endpoint (ex: /file)
--- a/tests/lib/box-client-test.js
+++ b/tests/lib/box-client-test.js
@@ -1163,6 +1163,19 @@ describe('box-client', function() {
 		});
 	});
 
+	describe('getToken()', function() {
+
+		it('should return an access token for the current client session', function() {
+			sandbox.mock(apiSessionFake).expects('getAccessToken')
+				.returns(Promise.resolve(FAKE_ACCESS_TOKEN));
+
+			return basicClient.getToken(null)
+				.then(data => {
+					assert.strictEqual(data, FAKE_ACCESS_TOKEN);
+				});
+		});
+	});
+
 	describe('plug()', function() {
 
 		var PLUGIN_API_ROOT = 'http://www.foobar.com',


boxcla:
Hi @lyaa, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

lyaa:
CLA signed

boxcla:
Verified that @lyaa has just signed the CLA. Thanks, and we look forward to your contribution.

mattwiller:
Hi @lyaa — thanks for the pull request!  This looks good, but I had two questions before we merge:

1. Can you please add a test for the new method?  We have a minimum code coverage level for the SDK, and your change caused CI to fail because it lowered our overall test coverage below that limit.
2. Can you elaborate a bit more about the use case for this method?  I want to make sure we understand how it will be used in case there are other solutions we might want to consider now or in the future.

lyaa:
Sure!

1. I have added a minimal test. Good enough?

2. I have been building an application that uses the box node SDK in combination with a bunch of custom and commercial APIs. All my users are App Users and the access tokens associated with their accounts were not accessible through the SDK itself. Based on my experience so far, the getToken function solves the following issues:
	* When prototyping, the box elements cannot be used without having access tokens.
	* Instantiating box client instances to perform single actions such as fetching a list of folders or uploading one text file is wasteful and adds up. My use case involved a lot of small operations that are chained together as promises. The process I had to go through each time I needed to interact with the box API was less readable and less efficient than what I can do now by using the tokens I got through the box client and stored for the small repeated operations.
	* There are functionalities implemented in the box API that are not in the SDK yet. For advanced uses, this function makes it possible to use HTTP calls when needed.

mattwiller:
@lyaa The test looks good!  I have some follow-up questions to make sure I'm understanding your use case correctly:

1. We don't recommend that you use the UI Elements with the fully-scoped token directly from the SDK client — you should make sure to use [Token Exchange](https://github.com/box/box-node-sdk/blob/master/docs/authentication.md#token-exchange) to get an access token with only the minimum permissions necessary to protect the security of your application.  

2. Can you elaborate a bit more on what waste you're seeing here?  I'd like to make sure that I fully understand what you're doing here so we can ensure that the SDK is well-optimized.  Is it possible for you to share a bit more about the operations you're doing or some sample code for how you're using the SDK?  Before/after would be super useful for us.

3. The SDK client object already allows you to make [custom API calls](https://github.com/box/box-node-sdk/blob/master/docs/client.md#custom-api-calls), and it will handle auth, retries, etc for you.  Is that different than what you're trying to do?  Also, what functionality are you trying to use that isn't in the SDK yet?  I'd be more than happy to ensure it gets added promptly!

4. It's not super well-documented (sorry — we're working on it!), but the SDK does provide a [way to just get the tokens for an app user](http://opensource.box.com/box-node-sdk/jsdoc/BoxSDKNode.html#getAppUserTokens) without instantiating a full client.  Would that solve your use case?

lyaa:
Thanks for the links @mattwiller. The method of getting the tokens for App Users looks promising. I don't have the time at the moment to test it out and answer your questions in details but I will take the time to do so in a couple of days.

I had used the UI Elements for prototyping so I wasn't worried about the scope of the access tokens. Right now, I handle all the access tokens for my users on the server side. The client-side only has authentication to my app and I send data over as needed.

PR #217, opened 2018-01-29
Add support for MDDR

mattwiller:


boxcla:

tests/lib/managers/retention-policies-test.js
```diff
@@ -326,6 +326,31 @@ describe('RetentionPolicies', function() {
...
+		it('should make POST request to assign policy with optional params when options are passed', function() {
+
+			var options = {
+				filter_fields: [
+					{
+						field: 'foo',
+						value: 'bar'
+					},
+					{
+						field: 'baz',
```
carycheng: Looks good! Can you specify more than one filter fields currently though? From the Note in the md

mattwiller:
@carycheng The API doesn't currently accept more than one filter field, but the SDK should be able to send more than one for forwards compatibility.

carycheng:
@mattwiller, gotcha! makes sense!

PR #222, opened 2018-02-07
Add support for streaming representation content

### LINKED ISSUE #180: Implement getReadStream for Representations
The current methods for fetching the representationInfo is great for checking the existence of the representations. If one does exist, there isn't a clean way to retrieve the representation content binary stream itself. Its definitely possible, but the developer experience is not ideal.

Current Experience: 
* Get Representation Info to check if there is a pdf, text, or image representation exists
* If one exists, get the content.url_template for the corresponding representation
* Trim the trailing text that contains the excess "{+asset_path}"
* Use the newly edited content url within the helper function client.get(pdf_rep_url_here)
* Get the binary content from the response body

Desired Experience:
* Get Representation Info to check if there is a pdf, text, or image representation exists
* Use a new function to get the stream of the representation client.files.getRepresentationStream('file_id', '[pdf]', callback)

mattwiller:
Implementing `files.getRepresentationContent()` to allow getting a read stream to the actual contents of a representation.  This will poll the representation info endpoint until the representation is generated.

Fixes #180


boxcla:

lib/managers/files.js
```diff
@@ -1344,6 +1375,53 @@ Files.prototype.getRepresentationInfo = function(fileID, representationType, cal
...
+ * @param {Object} [options] Optional parameters
+ * @param {string} [options.assetPath] Asset path for representations with multiple files
+ * @param {Function} [callback] Passed a stream over the representation contents if successful
+ * @returns {Promise<Readable>} A promise resolving to a stream over the representation contents
+ */
+Files.prototype.getRepresentationContent = function(fileID, representationType, options, callback) {
+
+	options = Object.assign({ assetPath: '' }, options);
+
+	return this.getRepresentationInfo(fileID, representationType)
```
ahmedcharles: This is perhaps better renamed to `getRepresentationEntries` (`Info` is confusing in the context of polling the info endpoint.) This is assuming that the rename isn't a breaking change.

lib/managers/files.js
```diff
@@ -1344,6 +1375,53 @@ Files.prototype.getRepresentationInfo = function(fileID, representationType, cal
...
+		.then(reps => {
+
+			var repInfo = reps.entries.pop();
```
ahmedcharles: I'm assuming you're expecting 0 or 1 entries because you expect `X-Rep-Hints` to only ask for a single thing? That should be mentioned in the function docs above, or better, it should be enforced somehow.

Note: my javascript isn't ideal, but I'm assuming `pop` removes the last element if there is one and returns javascript's null value if there is not. If that's the case, since the ordering of the returned list isn't specified, this is only deterministic for 0 or 1 entries.

lib/managers/files.js
```diff
@@ -1344,6 +1375,53 @@ Files.prototype.getRepresentationInfo = function(fileID, representationType, cal
...
+			if (!repInfo) {
+				throw new Error('Could not get information for requested representation');
+			}
+
+			switch (repInfo.status.state) {
```
ahmedcharles: Note, if `representationType` is the empty string (or indistinguishable from it to the server), the returned object won't have a `status` field and this will (may) fail?

lib/managers/files.js
```diff
@@ -93,6 +94,36 @@ function createFileContentFormData(content) {
...
+				throw errors.buildUnexpectedResponseError(response);
+			}
+
+			var info = response.body;
+
+			if (info.status.state === 'error') {
+				throw errors.buildResponseError(response, 'Representation had error status');
+			}
+
+			if (info.status.state !== 'success') {
```
ahmedcharles: This function should do the same thing as the switch statement below. Seems like a good place to extract a function and all it in both places. (I.e. this function doesn't check for `viewable`.

lib/managers/files.js
```diff
@@ -1344,6 +1375,53 @@ Files.prototype.getRepresentationInfo = function(fileID, representationType, cal
...
+			switch (repInfo.status.state) {
+
+			case 'success':
+			case 'viewable':
+				return repInfo.content.url_template;
+			case 'error':
+				throw new Error('Representation had error status');
+			case 'none':
+			case 'pending':
+				return pollRepresentationInfo(this.client, repInfo.info.url);
```
ahmedcharles: This should have a delay as well, not the other comment about extracting into a common function.

lib/managers/files.js
```diff
@@ -1344,6 +1375,53 @@ Files.prototype.getRepresentationInfo = function(fileID, representationType, cal
...
+			default:
+				throw new Error(`Unknown representation status: ${repInfo.status.state}`);
+			}
+		})
+		.then(assetURLTemplate => {
+			var url = assetURLTemplate.replace(ASSET_PATH, options.assetPath);
```
ahmedcharles: Ideally you would use something like: https://github.com/robhicks/tiny-url-template (assuming it actually implements the RFC correctly, I just did a quick search). Using replace results in us not being able to extend the url in the future without breaking clients. (note, the preview ui kit already uses replace, but I'd prefer to at least make the correct solution an option that is explicitly considered.)

lib/managers/files.js
```diff
@@ -1344,6 +1375,53 @@ Files.prototype.getRepresentationInfo = function(fileID, representationType, cal
...
+ * Method : GET
+ *
+ * @param {string} fileID The file ID to get the representation of
+ * @param {string} representationType The X-Rep-Hints type to request
+ * @param {Object} [options] Optional parameters
+ * @param {string} [options.assetPath] Asset path for representations with multiple files
+ * @param {Function} [callback] Passed a stream over the representation contents if successful
+ * @returns {Promise<Readable>} A promise resolving to a stream over the representation contents
+ */
+Files.prototype.getRepresentationContent = function(fileID, representationType, options, callback) {
```
ahmedcharles: It's possible to use a regular expression to validate that `representationType` is valid and will only ask for a single representation to be returned (i.e. guarantees 0 or 1 results). If regexes are easy enough in javascript and you'd like to do that, let me know and I'll give you a regex for it (once I figure out what syntax javascript uses).

tests/endpoint-test.js
```diff
@@ -1293,6 +1298,91 @@ describe('Endpoint', function() {
 				});
 			});
 		});
+
+		describe('getRepresentationContent()', function() {
```
ahmedcharles: I'm not entirely sure how the tests work, so I'll stop by and perhaps you can give me a high level overview of what's going on so I can just focus on the important parts.

lib/managers/files.js
mattwiller: We don't generally do any client-side validation to avoid needing to stay in lock-step with the API and to reduce our surface area for bugs (we don't want to accidentally reject valid input).  We treat the API as the source of truth for what input is valid and rely on it to return an error if the user passes invalid input.  All adding a regex would do is let the user avoid making an API call to see the error; I'm not sure it's worth it, especially since the X-Rep-Hints value is likely developer-provided.

lib/managers/files.js
mattwiller: This method is already live and cannot be renamed without a breaking change.  I don't think a simple rename is worth the hassle.

lib/managers/files.js
mattwiller: I think the non-determinism is acceptable — we can document around it.  I'm not especially keen on trying to do validation on their string input in the SDK.

lib/managers/files.js
mattwiller: That's a good point — I'll add basic sanity checking to ensure that a non-empty string did get passed.

lib/managers/files.js
mattwiller: I don't think we want to inject unnecessary latency here — if the status was `"none"`, we need to make the request to the info URL immediately to kick off generation of the representation.

lib/managers/files.js
mattwiller: Interesting, didn't realize this was a standard thing and not just our own replacement syntax.  I'll pull in a library to handle it.

lib/managers/files.js
ahmedcharles: Documentation is sufficient to assuage me. When you update the docs, just add me to the review/change, however that works.

lib/managers/files.js
ahmedcharles: That's fine.

lib/managers/files.js
```diff
@@ -93,6 +94,36 @@ function createFileContentFormData(content) {
...
+			}
+
+			var info = response.body;
+
+			if (info.status.state === 'error') {
+				throw errors.buildResponseError(response, 'Representation had error status');
+			}
+
+			if (info.status.state !== 'success') {
+				return Promise.delay(1000).then(() => pollRepresentationInfo(client, infoURL));
```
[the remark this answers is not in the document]
ahmedcharles: Sure.

lib/managers/files.js
ahmedcharles: Ah, yes. The delay is only beneficial if the status is pending, but it's not required. Thanks for pointing that out.

lib/managers/files.js
ahmedcharles: Thanks for being the first person to just accept my reasoning and pull in a library to do it. :)

lib/managers/files.js
ahmedcharles: That's fine, I just wanted to give the option.

PR #236, opened 2018-03-06
Add option to download byte range for file

### LINKED ISSUE #45: Support for Range Headers for files.getReadStream()
It was discovered in the API docs that the file download URL endpoint only supported version as a query string, while Range and BoxAPI is a header (Link: https://docs.box.com/reference#download-a-file) (initially listed Range, version, and BoxAPI all as acceptable query strings).

I'm wondering if the Node SDK will support the "Range" header as a param of the getReadStream() method, as "qs" was supposed to support the documented query strings from the API docs.

mattwiller:
Fixes #45

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -93,6 +93,20 @@ client.files.getReadStream('12345', null, function(error, stream) {
 });
 ```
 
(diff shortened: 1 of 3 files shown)


boxcla:

PR #246, opened 2018-03-17
Allow reconfiguring SDK instance root URL

mattwiller:
Certain SDK configuration properties need to be propagated to all
internal SDK objects (e.g. the TokenManager), which requires
re-instantiating those objects with the new Config object.  This
mostly affects widely-used configuration properties such as the
apiRootURL.


boxcla:

PR #251, opened 2018-03-22
Updated "files" documentation

paulomoreno:
After having issues when uploading a new file version to box, I noticed that the documentation had a small typo:

It's telling the developers to use the "uploadFile" function instead of the "uploadNewFileVersion" function. Using the wrong function causes an error.

I have signed the CLA.


boxcla:
Verified that @paulomoreno has signed the CLA. Thanks for the pull request!

mattwiller:
Good catch — thanks for the PR!

PR #253, opened 2018-03-23
Allow passing body Object to webhook message validator

### LINKED ISSUE #252: validateWebhookMessage - TypeError: Data must be string or buffer
### Description of the Issue
The crypto module was throwing a type error

### Versions Used
Node SDK: 1.14.1 
Node.js: // 8.10.0

### Steps to Reproduce
I made the following call to the box sdk, verifying that all values were non null and correct:

var isValid = BoxSDK.validateWebhookMessage(body, headers, primarySignatureKey, secondarySignatureKey);

The body and headers values were taken directly from an express res object. As a workaround, I modified my call like this: 

var isValid = BoxSDK.validateWebhookMessage(JSON.stringify(body), headers, primarySignatureKey, secondarySignatureKey);

I would think it would be common to have the body not be a string or buffer, and I was surprised to not see this error called out previously. If I've overlooked something obvious, please let me know.

### Error Message, Including Stack Trace
2018-03-22T18:16:09.797Z sm:indexjs TypeError: Data must be a string or a buffer
    at Hmac.update (crypto.js:99:16)
    at computeSignature (/usr/src/app/node_modules/box-node-sdk/lib/managers/webhooks.js:57:8)
    at validateSignature (/usr/src/app/node_modules/box-node-sdk/lib/managers/webhooks.js:80:28)
    at Function.Webhooks.validateMessage [as validateWebhookMessage] (/usr/src/app/node_modules/box-node-sdk/lib/managers/webhooks.js:337:8)
    at Object.boxHookChangeEvent (/usr/src/app/lib/services/box.js:161:24)
    at boxHook (/usr/src/app/routes/services/box.js:39:7)
    at Layer.handle [as handle_request] (/usr/src/app/node_modules/express/lib/router/layer.js:95:5)
    at next (/usr/src/app/node_modules/express/lib/router/route.js:131:13)
g.

[sdk-docs]: ./docs
[api-docs]: https://developer.box.com/docs

mattwiller:
Fixes #252

### DIFF
--- a/lib/managers/webhooks.js
+++ b/lib/managers/webhooks.js
@@ -311,7 +311,7 @@ Webhooks.setSignatureKeys = function(primaryKey, secondaryKey) {
 /**
  * Validate a webhook message by verifying the signature and the delivery timestamp
  *
- * @param {string} body - The request body of the webhook message
+ * @param {string|Object} body - The request body of the webhook message
  * @param {Object} headers - The request headers of the webhook message
  * @param {string} [primaryKey] - The primary signature to verify the message with. If it is sent as a parameter,
      it overrides the static variable primarySignatureKey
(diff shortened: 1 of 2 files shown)


boxcla:

PR #254, opened 2018-03-27
Improve error object format

mattwiller:
Adding to the standard error messages in the SDK to include common
fields necessary for troubleshooting, and attaching the request
object directly to the error when possible.


boxcla:

PR #256, opened 2018-03-31
Allow using Token Store with app auth client

mattwiller:
In order to allow caching of app auth tokens between clients or different processes, the TokenStore interface can now be used with app auth clients.  Since app auth permits having multiple different access tokens "live" at the same time, the logic for storing and invalidating tokens is much less strict than with traditional OAuth2.


boxcla:

Issue #276, opened 2018-05-16
Add an option to create a basic client without initializing a dummy


cbetta:
Currently, to initialize a BasicClient with an access token, you have to first initiallize a dummy SDK instance.

```js
let Box = require('box-node-sdk');
let box = new Box({
    clientID: 'unused',
    clientSecret: 'unused'
});
let client = box.getBasicClient(actualAccessToken);
```

This feels really redundant and ugly. Is there a chance we can simplify this by adding a "static" method to the BoxSDK class?

```js
let Box = require('box-node-sdk');
let client = Box.getBasicClient(actualAccessToken);
```

Under the hood, this could obviously do the same as the first code sample, but it would make all client code a lot cleaner and simpler.

mattwiller:
Closing since #278 is merged 🎉 — this will go out with the next release, tentatively scheduled for tomorrow

PR #277, opened 2018-05-16
Add getBasicClient alias to BoxSDKNode class

cbetta:
This allows for simpler creation of a basic client, without the need to instantiate a dummy SDK first. See #276 for more information.

```
let Box = require('box-node-sdk');
let client = Box.getBasicClient(actualAccessToken);
```

### DIFF
--- a/lib/box-node-sdk.js
+++ b/lib/box-node-sdk.js
@@ -195,10 +195,11 @@ BoxSDKNode.prototype.configure = function(params) {
  * @param {string} accessToken A user's Box API access token
  * @returns {BoxClient} Returns a new Box Client paired to a new BasicAPISession
  */
-BoxSDKNode.prototype.getBasicClient = function(accessToken) {
+BoxSDKNode.getBasicClient = function(accessToken) {
 	var apiSession = new BasicAPISession(accessToken, this.tokenManager);
 	return new BoxClient(apiSession, this.config, this.requestManager);
 };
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @cbetta, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

cbetta:
CLA signed

boxcla:
Verified that @cbetta has just signed the CLA. Thanks, and we look forward to your contribution.

lib/box-node-sdk.js
```diff
@@ -195,10 +195,11 @@ BoxSDKNode.prototype.configure = function(params) {
  * @param {string} accessToken A user's Box API access token
  * @returns {BoxClient} Returns a new Box Client paired to a new BasicAPISession
  */
-BoxSDKNode.prototype.getBasicClient = function(accessToken) {
+BoxSDKNode.getBasicClient = function(accessToken) {
 	var apiSession = new BasicAPISession(accessToken, this.tokenManager);
 	return new BoxClient(apiSession, this.config, this.requestManager);
```
mattwiller: I'm not sure this will work as expected — `this.config` and `this.requestManager` are instance properties of the SDK object, and won't be present in a static method like this.

cbetta: Ah yes, the test passed but not sure that tested the right thing. Let me try this again and actually test it properly :)

PR #278, opened 2018-05-22
Add class level getBasicClient

cbetta:
This allows for simpler creation of a basic client, without the need to instantiate a dummy SDK first. See #276 for more information.

```
let Box = require('box-node-sdk');
let client = Box.getBasicClient(actualAccessToken);
```

(Attempt #2 /cc @mattwiller)

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ Box supports four different types of client:
 Returns a Box Client with a Basic API Session. The client is able to make requests on behalf of a user. A basic session has no access to a user's refresh token. Because of this, once the session's tokens expire the client cannot recover and a new session will need to be generated.
 
 ```js
-var client = sdk.getBasicClient('ACCESS_TOKEN');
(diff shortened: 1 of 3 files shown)


boxcla:

cbetta:
One remaining question: is there a reason the instance level `getBasicClient` exists as well? If so, should this class level one maybe be called something else?

mattwiller:
@cbetta Thank you for the updated PR — this looks like it should work!  Unfortunately, your branch is failing CI due to lack of code coverage — would you mind adding a test to ensure that this method works correctly?  Something like [this existing test](https://github.com/box/box-node-sdk/blob/master/tests/integration-test.js#L58) would be great.

cbetta:
@mattwiller have updated with a test. Still getting errors as a few dependencies have security issues, which I don't think should be fixed in this PR. Happy to pull those upstream changes when they go live.

mattwiller:
@cbetta I just merged a fix for the security issues — I'll update your branch to re-run CI

PR #281, opened 2018-05-30
add fullInfo option to getRepresentationInfo()

### LINKED ISSUE #265: Add an option to getRepresentationInfo to start request

### Description of the Issue
In the API, getting a file's representation info returns an object called `status`.  If the status's state is `none`, then Box won't create the representation until you make a GET request to the url in the `info` object.  It'd be great if, when you run `getRepresentationInfo` function, if the status is `none`, it kicked off the process by making the request.

### Versions Used
Node SDK: 1.16.0
Node.js: 8.9.4
[sdk-docs]: ./docs
[api-docs]: https://developer.box.com/docs

sujaygarlanka:
Fixes #265

### DIFF
--- a/lib/managers/files.js
+++ b/lib/managers/files.js
@@ -1359,39 +1359,11 @@ Files.prototype.representation = {
  * @param {string} fileID - Box ID of the file being requested
  * @param {client.files.representation} representationType - The x-rep-hints value the application should create a
  *    representation for. This value can either come from FileRepresentationType enum or manually created
+ * @param {Object} [options] - Additional options. Can be left empty
+ * @param {Boolean} [options.fullInfo] - Set to true to return representation info where all states resolve to success. Default is false
  * @param {Function} [callback] - Passed an array of representaton objects if successful
  * @returns {Promise<Object>} A promise resolving to the representation response objects
 */
-// Files.prototype.getRepresentationInfo = function(fileID, representationType, callback) {
-//	var params = {
-//		qs: {
-//			fields: 'representations'
-//		},
-//		headers: {
-//			'x-rep-hints': representationType
-//		}
-//	};
-//	var apiPath = urlPath(BASE_PATH, fileID);
-//
-//	return this.client.get(apiPath, params)
-//		.then(response => {
-//			switch (response.statusCode) {
-//			// 202 - A Box file representation will be generated, but is not ready yet
-//			case httpStatusCodes.ACCEPTED:
-//				throw errors.buildResponseError(response, 'Representation not ready at this time');
-//
-//			// 200 - A Boxfile representation generated successfully
-//			// return the representation object
-//			case httpStatusCodes.OK:
-//				return response.body.representations;
-//
-//			// Unexpected Response
-//			default:
-//				throw errors.buildUnexpectedResponseError(response);
-//			}
-//		})
-//		.asCallback(callback);
-// };
 
 Files.prototype.getRepresentationInfo = function(fileID, representationType, options, callback) {
 	if (!representationType) {
--- a/lib/managers/files.js
+++ b/lib/managers/files.js
@@ -116,9 +116,8 @@ function pollRepresentationInfo(client, infoURL) {
 
 			case 'success':
 			case 'viewable':
-				return info;
 			case 'error':
-				throw errors.buildResponseError(response, 'Representation had error status');
+				return info;
 			case 'none':
 			case 'pending':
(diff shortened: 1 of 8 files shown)


boxcla:
Hi @sujaygarlanka, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

lib/managers/files.js
```diff
@@ -1359,10 +1359,20 @@ Files.prototype.representation = {
  * @param {string} fileID - Box ID of the file being requested
  * @param {client.files.representation} representationType - The x-rep-hints value the application should create a
  *    representation for. This value can either come from FileRepresentationType enum or manually created
+ * @param {Object} [options] - Additional options. Can be left empty
+ * @param {Boolean} [options.fullInfo] - Set to true to return representation info where all states resolve to success. Default is false
  * @param {Function} [callback] - Passed an array of representaton objects if successful
  * @returns {Promise<Object>} A promise resolving to the representation response objects
 */
-Files.prototype.getRepresentationInfo = function(fileID, representationType, callback) {
+
```
mattwiller: This newline between the JSDoc comment and the method should be removed

lib/managers/files.js
```diff
@@ -1359,10 +1359,20 @@ Files.prototype.representation = {
...
+Files.prototype.getRepresentationInfo = function(fileID, representationType, options, callback) {
+	if (!representationType) {
```
mattwiller: This check is likely not backwards-compatible, since this method would previously worked without specifying X-Rep-Hints.  You'll probably need to add checks later when handling the response to ensure that your code works correctly when no X-Rep-Hints are passed

lib/managers/files.js
```diff
@@ -1383,6 +1393,21 @@ Files.prototype.getRepresentationInfo = function(fileID, representationType, cal
 			// 200 - A Boxfile representation generated successfully
 			// return the representation object
 			case httpStatusCodes.OK:
+
+				if (options && options.fullInfo) {
+
+					var data = response.body.representations.entries;
+					var promiseArray = data.map(entry => {
+						if (entry.status.state === 'success' || entry.status.state === 'viewable') {
```
mattwiller: This should probably also include the `"error"` state, which isn't something we should poll on.  Might be worth adding a small test where one of the reps is errored and define the behavior for that case.

tests/endpoint-test.js
```diff
@@ -1299,6 +1299,64 @@ describe('Endpoint', function() {
...
 		});
 
+		describe('getRepresentationInfo()', function() {
+
+			it('should resolve all representation entries with success status when fullInfo option is passed',
+				function() {
+
+					var repsFixtureMultiple = getFixture('files/get_files_id_representations_multiple_200'),
+						repsObj = JSON.parse(repsFixtureMultiple),
+						//						repPNGURL = url.parse(repsObj.entries[0].info.url).pathname,
```
mattwiller: This commented-out line can be removed

lib/managers/files.js
```diff
@@ -1359,10 +1359,19 @@ Files.prototype.representation = {
  * @param {string} fileID - Box ID of the file being requested
  * @param {client.files.representation} representationType - The x-rep-hints value the application should create a
  *    representation for. This value can either come from FileRepresentationType enum or manually created
+ * @param {Object} [options] - Additional options. Can be left empty
+ * @param {Boolean} [options.generateRepresentations] - Set to true to return representation info where all states resolve to success. Default is false
```
mattwiller: The type for this should be `{bool}` instead of `{Boolean}` — we expect it to be the primitive, not the object type.  You can also indicate the default value like this: `[options.generateRepresentations=false]`

lib/managers/files.js
```diff
@@ -1431,7 +1461,14 @@ Files.prototype.getRepresentationContent = function(fileID, representationType,
...
 			case 'pending':
-				return pollRepresentationInfo(this.client, repInfo.info.url);
+				return pollRepresentationInfo(this.client, repInfo.info.url)
+					.then(info => {
+						if (info.status.state === 'error') {
+							throw new Error('Representation had error status');
+						}
+						return info;
+					})
+					.then(info => info.content.url_template );
```
mattwiller: You can combine these two `.then()`s for better performance:

```js
.then(info => {
    if (info.status.state === 'error') {
        throw new Error('Representation had error status');
    }
    return info.content.url_template;
})
```

PR #303, opened 2018-08-07
Remove nsp check

mattwiller:
The Node Security Project will be shutting down next month, and
will be replaced by checks within npm.  At this time, we can
safely remove the check from our own test command.


boxcla:

PR #304, opened 2018-08-07
destroyable EventStreams; closes #302

### LINKED ISSUE #302: event stream destruction
There doesn't seem to be any way to *stop* an `EventStream` from continually long-polling once started, short of killing the Node.js process.

In Node.js v8.x or later I think you want to implement the `Readable` stream's `_destroy()` (much like you would with `_read()`) method.

For Node.js pre-v8.x, generally it looks like you're going to want to *wrap* `Readable#destroy()`.  So a solution might look like this:

```js
EventStream.prototype._destroy = function() {
  // if this is v8.x+ and `EventStream#destroy()` was called, then 
  // `this.destroyed` will be true.  if it isn't, then call it,
  // since we probably got here via `MyStream#abort()`.
  // `Readable#destroy()` sets `this.destroyed` to `true`.
  if (!this.destroyed) {
    this.destroy();
  }
  // shut down long polling
});

EventStream.prototype.abort = EventStream.prototype._destroy;
```

My use-case:  I'm implementing event streams in a [Node-RED](https://nodered.org) addon, which will otherwise leak memory and resources.

Hopefully this sounds OK; I have an implementation and tests which I'll send as soon as tomorrow.

boneskull:
- add reference to long-polling timer on class
- add `EventStream` method `_destroy()`, which clears any existing
  long-polling timer, and deletes it. Destroys stream if not already
  destroyed.  Implements `Readable#_destroy` in Node.js v8.x and newer.
- add `EventStream` method `abort()`, an alias for `_destroy()`, which
  is *not* called via `EventStream#destroy()` in pre-Node.js-v8.x.  Any
  version can use this method; `_destroy()` is private.
- add checks for `destroyed` stream state which effectively abort
  various tasks in `EventStream`
- consolidate long-polling retry logic into its own private method,
  `EventStream#retryLongPoll()`.
- add tests for logic changes
- add note in docs

* * *

It'd be helpful to have an integration test (*without* use of Sinon's fake timers) which asserts a destroyed stream doesn't still have tasks in the event loop.  If it does, such a test should hang--ostensibly due to the infinite recursion--because of Mocha's v4.x+ behavior.

Originally, I discovered this problem when writing some test code which consumes an `EventStream`.  I could determine that *this* implementation was correct, because my test code was no longer hanging indefinitely.  :smile:


boxcla:
Verified that @boneskull has signed the CLA. Thanks for the pull request!

mattwiller:
@boneskull Overall the implementation looks good to me, but it looks like tests aren't passing on Node.js versions before 8.x — you'll need to ensure your code works correctly for versions all the way back to Node.js 4.0.0

boneskull:
The coverage % decreased because this PR adds more covered LoC to the project.  I don't love how Coveralls calculates that...

boneskull:
@mattwiller The failures have been fixed on Travis-CI

lib/event-stream.js
```diff
@@ -269,4 +296,38 @@ EventStream.prototype._read = function() {
...
+ * `destroy()`.
+ * @returns {void}
+ * @private
+ */
+EventStream.prototype._destroy = function() {
+
+	if (!this.destroyed) {
+		/* istanbul ignore else */
+		if (typeof this.destroy === 'function') {
+			this.destroy();
```
mattwiller: The [Node.js Stream docs](https://nodejs.org/api/stream.html#stream_readable_destroy_err_callback) say that `_destroy()` is called by `destroy()` — why call back into `destroy()` here?

lib/event-stream.js
```diff
@@ -269,4 +296,38 @@ EventStream.prototype._read = function() {
...
+	delete this._retryTimer;
+};
+
+/**
+ * Destroys this stream; for Node.js v8.0.0 and newer, just use
+ * `EventStream#destroy()`.
+ * @returns {void}
+ * @public
+ */
+EventStream.prototype.abort = EventStream.prototype._destroy;
```
mattwiller: It's a bit undesirable to have two different methods to call depending on which version of Node.js the user is running.  Is it possible to check if `EventStream.prototype.destroy` already exists and conditionally add/polyfill it if it doesn't?

lib/event-stream.js
boneskull: Because `destroy` was not added until v8 of Node.

lib/event-stream.js
boneskull: yeah, we could do that.

lib/event-stream.js
boneskull: (so if you called `abort` you might as well have called `destroy` if it exists. this is moot if I just conditionally add the method)

boneskull:
@mattwiller Build needs retry; Node.js v9 didn't seem to start.

lib/event-stream.js
```diff
@@ -269,4 +296,34 @@ EventStream.prototype._read = function() {
...
+ * @returns {void}
+ * @private
+ */
+EventStream.prototype._destroy = function() {
+	clearTimeout(this._retryTimer);
+	delete this._retryTimer;
+};
+
+// backwards-compat for Node.js pre-v8.0.0
+if (!(typeof Readable.destroy === 'function')) {
```
mattwiller: You'll need to check against `Readable.prototype.destroy` here — this condition as written will always include the polyfill.  Also, using `!==` might be more readable than `!(... === ...)`

boneskull: my bad

boneskull:
@mattwiller addressed your latest comments

boneskull:
@mattwiller Do you need anything else from me here?

boneskull:
@mattwiller Possible to release this soon?

mattwiller:
@boneskull We're looking to push out a release later this week!

PR #309, opened 2018-08-23
Improve error messages for app authorization failures

### LINKED ISSUE #307: Thrown messages for some API errors may need improvement

### Description of the Issue

I was getting `Unexpected API Response [400 Bad Request]` in my enterprise app and could not figure out whether it was a problem with a JSON config, the server or something else. Here is my code:

```ts
const sdk = BoxSDK.getPreconfiguredInstance(boxConfig);
const client = sdk.getAppAuthClient("user", userId);
const items = await client.folders.getItems(boxFolderId);
```

Here's the look of the error stack:

```
Unexpected API Response [400 Bad Request]
    at Object.buildResponseError (/path/to/node_modules/box-node-sdk/lib/util/errors.js:96:23)
    at Object.buildUnexpectedResponseError (/path/to/node_modules/box-node-sdk/lib/util/errors.js:130:15)
    at requestManager.makeRequest.then.response (/path/to/node_modules/box-node-sdk/lib/token-manager.js:226:19)
    at tryCatcher (/path/to/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/path/to/node_modules/bluebird/js/release/promise.js:512:31)
    ...
```

Only after catching the error and adding `console.log(error.response.body)`, I saw: 

```js
{ error: 'unauthorized_client',
  error_description: 'This app is not authorized by the enterprise admin' }
```

It'd be great if [`lib/util/errors.js`](https://github.com/box/box-node-sdk/blob/master/lib/util/errors.js) was improved a bit and so that the stack trace became more informative by default. Just adding `console.log(error)` prints a massive object with enormous amounts of data and it is pretty hard to pinpoint the cause.

### Versions Used
Node SDK: 1.20.0
Node.js: 10.9.0

mattwiller:
Fixes #307

### DIFF
--- a/lib/util/errors.js
+++ b/lib/util/errors.js
@@ -82,12 +82,21 @@ module.exports = {
 		}
 
 		var apiMessage = '';
-		if (response.body && response.body.code) {
-			apiMessage += ` ${response.body.code}`;
-		}
+		// if (response.body && response.body.code) {
+		// 	apiMessage += ` ${response.body.code}`;
+		// }
 
-		if (response.body && response.body.message) {
-			apiMessage += ` - ${response.body.message}`;
+		// if (response.body && response.body.message) {
+		// 	apiMessage += ` - ${response.body.message}`;
+		// }
+
+		if (response.body) {
+
+			var code = response.body.code || response.body.error;
+			var description = response.body.message || response.body.error_description;
+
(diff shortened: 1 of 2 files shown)


boxcla:

PR #314, opened 2018-09-14
Chunked uploader improvements

mattwiller:
- Return a promise from `uploader.start()` and `uploader.abort()` to simplify use of the basic Chunked Upload operations
- Allow passing file attributes through the chunked uploader options


boxcla:

PR #327, opened 2018-11-29
Update description and keywords for SEO

cbetta:
Added keywords and updated the title to add in the words "Platform" and "API". Removed the "for Node" as that's self-explanatory.


boxcla:

PR #337, opened 2019-01-18
Allow reconfiguring of request retry strategy

buhrianlee:
Adds `config.retryStrategy` which is an optional config param. It is a function that takes in an object of options including the error, # of retries made so far, and the `config.numMaxRetries` and `config.retryIntervalMS`.  If the function returns a number, then this will be used as the time (in ms) until the next retry. If the function returns an Error, this will be propagated in the response to the user. If neither of those are returned, the original response error object will be propagated.

### DIFF
--- a/lib/api-request.js
+++ b/lib/api-request.js
@@ -285,8 +285,27 @@ APIRequest.prototype._retry = function(err) {
 	this.numRetries = this.numRetries || 0;
 
 	if (this.numRetries < this.config.numMaxRetries) {
+		var retryTimeout;
 		this.numRetries += 1;
-		setTimeout(this.execute.bind(this), getRetryTimeout(this.numRetries, this.config.retryIntervalMS));
+		if (this.config.retryStrategy) {
+			var retryOptions = {
+				error: err,
+				numRetries: this.numRetries,
+				numMaxRetries: this.config.numMaxRetries,
+				retryIntervalMS: this.config.retryIntervalMS
+			};
+			retryTimeout = this.config.retryStrategy(retryOptions);
+			if (typeof retryTimeout !== 'number') {
+				if (retryTimeout instanceof Error) {
+					err = retryTimeout;
+				}
+				this._finish(err);
+				return;
+			}
+		} else {
+			retryTimeout = getRetryTimeout(this.numRetries, this.config.retryIntervalMS);
+		}
+		setTimeout(this.execute.bind(this), retryTimeout);
 	} else {
 		err.maxRetriesExceeded = true;
 		this._finish(err);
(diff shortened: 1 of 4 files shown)


boxcla:
Hi @buhrianlee, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

buhrianlee:
CLA signed

boxcla:
Verified that @buhrianlee has just signed the CLA. Thanks, and we look forward to your contribution.

lib/api-request.js
```diff
@@ -285,8 +285,27 @@ APIRequest.prototype._retry = function(err) {
...
 	if (this.numRetries < this.config.numMaxRetries) {
+		var retryTimeout;
 		this.numRetries += 1;
-		setTimeout(this.execute.bind(this), getRetryTimeout(this.numRetries, this.config.retryIntervalMS));
+		if (this.config.retryStrategy) {
+			var retryOptions = {
+				error: err,
+				numRetries: this.numRetries,
+				numMaxRetries: this.config.numMaxRetries,
+				retryIntervalMS: this.config.retryIntervalMS
```
mattwiller: We should definitely add the number of attempts so far here, and also the total number of ms spent waiting if possible

buhrianlee:
@mattwiller addressed your comment and added integration test

tests/integration-test.js
```diff
@@ -218,6 +218,49 @@ describe('Box Node SDK', function() {
...
+				id: 'me'
+			});
+
+
+		var sdk = new BoxSDK({
+			clientID: TEST_CLIENT_ID,
+			clientSecret: TEST_CLIENT_SECRET,
+			maxNumRetries: 5,
+			retryIntervalMS: 10,
+			retryStrategy(options) {
```
mattwiller: It might be nice to make the `retryStrategy` a `sinon` stub, so that we can assert that it was called the expected number of times as well, and with the correct arguments.

tests/integration-test.js
```diff
@@ -218,6 +218,64 @@ describe('Box Node SDK', function() {
...
+				// Retry strategy function should be passed arg object with correct properties
+				var args = retryStrategyStub.getCall(0).args;
+				assert.lengthOf(args, 1);
+				assert.hasAllKeys(args[0], [
+					'error',
+					'numMaxRetries',
+					'numRetryAttempts',
+					'retryIntervalMS',
+					'totalElapsedTimeMS'
+				]);
```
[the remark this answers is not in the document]
buhrianlee: Done. For `totalElapsedTimeMS`, I've just added an assertion that it is a number since that param's value is hard to predict.

PR #356, opened 2019-02-23
Add trace ID to API error message

mattwiller:
The Box API now sends back a second request ID (trace ID) in the
BOX-REQUEST-ID header.  When this value is present, it will be
included in SDK response error messages to aid in troubleshooting.

### DIFF
--- a/lib/util/errors.js
+++ b/lib/util/errors.js
@@ -10,7 +10,7 @@
 var qs = require('querystring'),
 	httpStatusCodes = require('http-status');
 
-var TRACE_ID_HEADER_NAME = 'box-request-id';
+const TRACE_ID_HEADER_NAME = 'box-request-id';
 
 // ------------------------------------------------------------------------------
 // Typedefs and Callbacks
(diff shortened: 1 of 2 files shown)


boxcla:

boxcla:

lib/util/errors.js
```diff
@@ -10,6 +10,7 @@
 var qs = require('querystring'),
 	httpStatusCodes = require('http-status');
 
+var TRACE_ID_HEADER = 'box-request-id';
```
nidoran:
```suggestion
const TRACE_ID_HEADER_NAME = 'box-request-id';
```

lib/util/errors.js
```diff
@@ -10,6 +10,7 @@
 var qs = require('querystring'),
 	httpStatusCodes = require('http-status');
 
+var TRACE_ID_HEADER_NAME = 'box-request-id';
```
nidoran: const?

PR #380, opened 2019-04-24
Tag code samples for usage in developer docs

cbetta:
@mattwiller I've tagged the majority of code samples and added a few more code samples to help showcase some of the authentication endpoints.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -21,6 +21,7 @@ overview of how the Box API handles authentication.
 - [As-User](#as-user)
 - [Token Exchange](#token-exchange)
     - [Annotator Tokens](#annotator-tokens)
+- [Revoking Tokens](#revoking-tokens)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -320,15 +321,16 @@ client.exchangeToken('item_preview', 'https://api.box.com/2.0/files/123456', opt
 This will attach an external user name and ID to annotations made with the token,
 in order to attribute them to someone who does not have a Box account.
 
-### Revoking tokens
+## Revoking tokens
 
-The access tokens for a client can be revoked when needed.
+Access tokens for a client can be revoked when needed. As this removes the client's way of authenticating this client can no
+longer be used after this call.
 
 <!-- sample post_oauth2_revoke -->
 ```js
 client.revokeTokens()
 	.then(() => {
-		// the access token has been revoked
+		// the client's access token have been revoked
 	});
 ```
 
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -120,21 +120,29 @@ existing Box users, to integrate with their accounts.
 Your application must provide a way for the user to login to Box (usually with a
 browser or web view) in order to obtain an auth code.
 
+<!-- sample get_authorize -->
+```js
+var BoxSDK = require('box-node-sdk');
+var sdk = new BoxSDK({
+	clientID: 'YOUR-CLIENT-ID',
+	clientSecret: 'YOUR-CLIENT_SECRET'
+});
+
+// the URL to redirect the user to
+var authorize_url = BoxSDK.getAuthorizeURL({
+	response_type: 'code'
+});
+```
+
 After a user logs in and grants your application access to their Box account,
 they will be redirected to your application's `redirect_uri` which will contain
 an auth code. This auth code can then be used along with your client ID and
 client secret to establish an API connection.  A `PersistentClient` will
 automatically refresh the access token as needed.
 
+<!-- sample post_oauth2_token --->
 ```js
-var BoxSDK = require('box-node-sdk');
-var TokenStore = require('TOKEN-STORE-IMPLEMENTATION');
-var sdk = new BoxSDK({
-	clientID: 'YOUR-CLIENT-ID',
-	clientSecret: 'YOUR-CLIENT_SECRET'
-});
-
-sdk.getTokensAuthorizationCodeGrant('YOUR-AUTH-CODE', null, function(err, tokenInfo) {
+sdk.getTokensAuthorizationCodeGrant('<CODE>', null, function(err, tokenInfo) {
 
 	if (err) {
 		// handle error
@@ -311,3 +319,16 @@ client.exchangeToken('item_preview', 'https://api.box.com/2.0/files/123456', opt
 
 This will attach an external user name and ID to annotations made with the token,
 in order to attribute them to someone who does not have a Box account.
+
+### Revoking tokens
+
+The access tokens for a client can be revoked when needed.
+
+<!-- sample post_oauth2_revoke -->
+```js
+client.revokeTokens()
+	.then(() => {
+		// the access token has been revoked
+	});
+```
+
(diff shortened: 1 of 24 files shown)


boxcla:

docs/metadata.md
```diff
@@ -492,6 +503,7 @@ Metadata can be created on a folder by calling
 [`folders.addMetadata(folderID, scope, template, metadata, callback)`](http://opensource.box.com/box-node-sdk/jsdoc/Folders.html#addMetadata)
 with a metadata template and an object of key/value pairs to add as metadata.
 
+<!-- sample post_folders_id_metadata_id_id -->
```
mattwiller: We're in the process of adding a convenience method for metadata that tries the POST and falls back to PUT if there's a Conflict error; we would love to steer the vast majority of developers to use this method since it guarantees application of the provided key/value pairs.  How should we handle cases like this, where the SDK provides both a (preferred) convenience method, and a low-level method that maps 1:1 to the endpoint?

docs/storage-policies.md
```diff
@@ -100,6 +104,7 @@ To determine which storage policy is assigned to a user, call
 [`storagePolicies.getAssignmentForTarget(userID, callback)`][getAssignmentForTarget]
 with the ID of the user.
 
+<!-- sample get_storage_policy_assignments -->
```
mattwiller: API WG has been promoting a pattern where related objects (e.g. the assignment for a user) are accessed via a query parameter on the related collection resource (`GET /storage_policy_assignments?user_id=ID`) rather than as a subresource on the individual resource (e.g. `GET /users/ID/storage_policy_assignment`, which does not exist).  This could be confusing to document if the collection resource should also be used more generally to get all of the elements in the collection (e.g. just `GET /storage_policy_assignments`, which is not used today for this endpoint) — any idea how we should tag those going forward, since theoretically they correspond to different logical operations?

docs/trash.md
```diff
@@ -239,6 +245,7 @@ Get a Trashed Folder
 
 Information about a folder in the trash can be retrieved with the [`folders.getTrashedFolder(folderID, options, callback)`](http://opensource.box.com/box-node-sdk/jsdoc/Folders.html#getTrashedFolder) method.
 
+<!-- sample get_folder_id_trash -->
```
mattwiller: Should be `get_folders_id_trash`

docs/users.md
```diff
@@ -245,6 +253,7 @@ Delete Email Alias
 
 To delete a users email alias call the [`users.removeEmailAlias(userID, aliasID, callback)`](http://opensource.box.com/box-node-sdk/jsdoc/Users.html#removeEmailAlias) method.
 
+<!-- sample delete_users_id_email_aliases -->
```
mattwiller: Should be `delete_users_id_email_aliases_id`

docs/authentication.md
```diff
@@ -311,3 +319,16 @@ client.exchangeToken('item_preview', 'https://api.box.com/2.0/files/123456', opt
 
 This will attach an external user name and ID to annotations made with the token,
 in order to attribute them to someone who does not have a Box account.
+
+### Revoking tokens
+
+The access tokens for a client can be revoked when needed.
```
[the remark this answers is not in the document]
cbetta: @mattwiller agreed. I'm thinking though, is there a different SDK api call that more directly relates to `POST /oauth2/revoke` without that side effect?

docs/metadata.md
```diff
@@ -247,6 +251,7 @@ Get all metadata templates for the current enterprise and scope by calling the
 [`metadata.getTemplates(scope, callback)`](http://opensource.box.com/box-node-sdk/jsdoc/Metadata.html#getTemplates)
 method.
 
+<!-- sample get_metadata_templates_enterprise -->
```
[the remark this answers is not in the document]
cbetta: So going off https://developer.box.com/reference#get-enterprise-templates it mentions only `enterprise` is supported.

Additionally, `get_metadata_templates_id` seems to map to https://developer.box.com/reference#get-metadata-template-by-id instead, which fetched a template by template ID.

docs/metadata.md
cbetta: I think this goes into the variations discussion we talked about a while ago. I think we can resolve that as we ship it. Happy to discuss once this goes live.

cbetta: Just to clarify, we could have samples like this:

Standard sample:
`<!-- sample post_folders_id_metadata_id_id -->`

Variation sample:
`<!-- sample post_folders_id_metadata_id_id with_fallback -->`

We can then show both samples, or only the fallback one in the ref docs.

docs/storage-policies.md
cbetta: Interesting. I'd prefer the alternative actually. I think of endpoints as unix commands: do one thing and do it well. 

As for this issue, ideally, I'd like the ref docs to just document all params for an endpoint and then have good guides, recipes, etc that showcase each use case for each endpoint. Yet, we can explore different ways to highlight usage of endpoints as we progress. This is already a thing we have to explore with some of the existing endpoints anyway.

As for the samples, I'd love if we can at least expose one method per SDK that matches closely to each endpoint. We can then handle the variations and clever functions in a different way in variation samples.

docs/metadata.md
cbetta: I've tagged the old API calls, for now.

cbetta:
Please hold off on merging.

docs/authentication.md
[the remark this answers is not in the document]
mattwiller: @cbetta You could add a second code sample showing how to [revoke a token directly](http://opensource.box.com/box-node-sdk/jsdoc/BoxSDKNode.html#revokeTokens) and make that one the tagged sample.  But I think for most users, `client.revokeTokens()` will be the main way they will revoke using the SDK, so that makes sense to me to keep.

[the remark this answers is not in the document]
cbetta: Yes and no. I'd like to document as close as possible how to use the endpoint in the reference docs. 

Am I right to think the sample should be?

```js
sdk.revokeTokens("<TOKEN>")
  .then(() => {
    // the token has been revoked
  })
```

docs/metadata.md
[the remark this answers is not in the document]
cbetta: I need to follow up and see what the case here is exactly.

cbetta:
Good for me to have the review continue.

docs/authentication.md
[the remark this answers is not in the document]
mattwiller: @cbetta Yep, that looks right to me!

docs/collaborations.md
```diff
@@ -346,6 +350,7 @@ Get Pending Collaborations
 A collection of all the user's pending collaborations can be retrieved with
 [`collaborations.getPending(callback)`](http://opensource.box.com/box-node-sdk/jsdoc/Collaborations.html#getPending).
 
+<!-- sample get_collaborations -->
```
mattwiller: Since this is a pretty specific case, we might want to make this one a variant, e.g. `get_collaborations pending`

cbetta: Not sure. I think that endpoint only returns pending collaborations.

cbetta:
🎉

PR #389, opened 2019-06-19
Split get_folders_id sample

cbetta:
This sample had two samples within the code block, which was a problem for the dev docs.

I've updated it to a generic sample, followed by a sample showing how to limit fields.

I also turned this second sample into one using promises, as we don't seem to use callbacks in any of the other samples.


boxcla:

PR #396, opened 2019-07-11
Add support for token exchange using shared links

naeims:


boxcla:
Verified that @naeims has signed the CLA. Thanks for the pull request!

PR #405, opened 2019-08-19
Regenerate JWT ID and retry auth request when JTI claim is rejected

### LINKED ISSUE #403: JTI claim value is not getting regenerated if any API calls fails due to rate limit

### Description of the Issue
We are using box-node-sdk v 1.29.0 in our application. The application is designed to discover all the files inside a folder and download the files found inside it. To achieve this, we to first API call to get the list of files inside the given folder and for each of the file discovered, we do one more API call to get details.

The code looks like below:

```
    var session = boxNodeSDK.getPreconfiguredInstance(this.getDataSource().connector.OAuth2_JWT_config);
    var serviceAccountClient = session.getAppAuthClient('user', userId);
    serviceAccountClient.files.get(fileId, {}, function(err, fileInfo) {
```
we observed that above code fails randomly with below error. upon reading further on this error, we are of understanding that this is due to rate limits being hit inside SDK.

```
{"error":"invalid_grant","error_description":"Please check the 'jti' claim. A unique 'jti' value is required."}
```

Could someone please help us understand if there is any way we can handle this condition using a retry strategy.

### Versions Used
Node SDK: 1.29.0
Node.js: 8.15.1

### Steps to Reproduce
The steps to reproduce would be to have an application making API calls at a very fast pace to Box.

### Error Message, Including Stack Trace
```
{"error":"invalid_grant","error_description":"Please check the 'jti' claim. A unique 'jti' value is required."}
```

mattwiller:
Fixes #403

### DIFF
--- a/lib/token-manager.js
+++ b/lib/token-manager.js
@@ -40,6 +40,21 @@
  *                                   refresh token, and have no refresh capabilities.
  */
 
+/**
+  *	Determines whether a JWT auth error can be retried
+  * @param {Error} err The JWT auth error
+  * @returns {boolean} True if the error is retryable
+  */
+function isJWTAuthErrorRetryable(err) {
+
+	return err.authExpired
+		&& err.response.headers.date
+		&& (
+			err.response.body.error_description.indexOf('exp') > -1
+				|| err.response.body.error_description.indexOf('jti') > -1
+		);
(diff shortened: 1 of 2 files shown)


boxcla:

PR #423, opened 2019-09-23
Update webhooks.md

cbetta:
Add a sample tag for verifying signatures.

Decided to use a new syntax of `x_tagname variant` for a SDK feature that does not directly match an endpoint. In this case, `x_webhooks validate_signatures` signifies the `validate_signatures` sample for the `webhooks` category.

This can be used in a guide, but won't risk appearing in the API reference.


boxcla:

nidoran:
Overriding coveralls b/c of documentation-only change

PR #445, opened 2019-10-24
Update authentication.md

cbetta:
Fixing sample to point to right variable to use to get authorize URL.


boxcla:

PR #461, opened 2020-01-31
Fix JWT retry logic

sujaygarlanka:


boxcla:
Verified that @sujaygarlanka has signed the CLA. Thanks for the pull request!

lib/token-manager.js
```diff
@@ -345,37 +349,92 @@ TokenManager.prototype = {
...
+				if (typeof retryTimeout !== 'number') {
+					if (retryTimeout instanceof Error) {
+						error = retryTimeout;
 					}
-
-					return this.getTokens(params, options);
+					throw error;
 				}
+			} else if (error.response.headers.hasOwnProperty('retry-after')) {
+				retryTimeout = error.response.headers['retry-after'] * 1000;
```
PJSimon: Should you add the retry-after check here, as well?
https://github.com/box/box-node-sdk/blob/master/lib/api-request.js#L321

tests/lib/token-manager-test.js
```diff
@@ -630,7 +629,181 @@ describe('token-manager', function() {
...
+		}).timeout(5000);
+
+		it('should retry with new jti claim when server rejects because of rate limiting', function() {
+
+			var firstTokenParams = {
+				grant_type: 'urn:ietf:params:oauth:grant-type:jwt-bearer',
+				assertion: TEST_WEB_TOKEN
+			};
+
+			var newJWT = 'jhdhioeiu4yo34875twudgshdgr';
```
PJSimon: How was this generated?  It should be a freshly-generated key from a non-Box account...

tests/lib/token-manager-test.js
```diff
@@ -630,7 +629,181 @@ describe('token-manager', function() {
...
+			var secondTokenParams = {
+				grant_type: 'urn:ietf:params:oauth:grant-type:jwt-bearer',
+				assertion: newJWT
+			};
+
+			var newJWT2 = 'jhighreiu4yo34875tfreggshdgr';
```
PJSimon: How was this generated?  It should be a freshly-generated key from a non-Box account...

tests/lib/token-manager-test.js
```diff
@@ -630,7 +629,181 @@ describe('token-manager', function() {
...
+					},
+					headers: {
+						'retry-after': 2,
+						date: serverDate2
+					}
+				}
+			};
+
+			var tokenInfo = {
+				accessToken: 'lsdjhgo87w3h4tbd87fg54'
```
PJSimon: How was this generated? It should be a freshly-generated key from a non-Box account...

sujaygarlanka: This is taken from other tests in the file.

tests/lib/token-manager-test.js
sujaygarlanka: This is taken from other tests in the file.

tests/lib/token-manager-test.js
sujaygarlanka: I just changed up some characters in the newJWT parameter.

Issue #463, opened 2020-02-03
files.uploadFile() not working with custom stream


F145h:
Have a generated text content on heroku(without write access to the filesystem), upload to box.com failed with error "Unexpected API Response [400 Bad Request] bad_request - Stream ended unexpectedly"

sample:
```
                const readable = new Readable();
                readable._read = () => {}

                readable.push(Buffer.from("Hello", 'utf8'));
                readable.push(null);

                boxConnection.files.uploadFile("0", "test.txt", readable).then(file => {
                     ....
                }
```

1) getChunkedUploader() works
2) empty stream( without `readable.push(Buffer.from("Hello", 'utf8'));`) works too
3) readable.readableLength has the correct value 5 bytes

sgalizia:
This is because Box uses `request` under the hood and the library requires that you provide a `content-length` header. On the Box community site they suggest editing the prototype of the stream to add these properties that `request` is looking for. I **STRONGLY** disagree with this. It is an underlying fault in their SDK and they need to provide a way for users to tell the SDK how big the file is that they want to upload.

I'm hoping to get their attention on this to fix it as I ran into the EXACT same issue yesterday and wasted hours trying to figure this out. When you are not using a `fs` stream it does not work. Give us the ability to use non `fs` streams and specify the size please!

F145h:
A good solution: make POST request for `https://upload.box.com/api/2.0/files/content` and upload content as multupart. Then you need to work and updating accessToken yourself. And SDK is not needed 🤨

PJSimon:
Hi @F145h and @sgalizia,

Thanks for bringing up this excellent feedback.  I'll discuss with the team and get back to you soon!

Patrick

sujaygarlanka:
Hello @F145h and @sgalizia,

What is the reason you guys have for not using fs streams? We would like to know so we can appropriately prioritize this request.

F145h:
> Hello @F145h and @sgalizia,
> 
> What is the reason you both have for not using fs streams? We would like to know so we can appropriately prioritize this request.

In my case heroku hosting does not allow write to the local file system, therefore I can’t use fs-only streams.
Well, custom stream does not work in the example as I showed, it is simple. But it works fine with dropbox, aws-sdk etc. 
Why do we need save a temporary file if the data is already have in the buffer / stream

sgalizia:
> 
> 
> Hello @F145h and @sgalizia,
> 
[...]

Issue #143, opened 2017-06-08
Cannot upload files from streams other than fs.createReadStream


joedski:
I've been trying to build a server which uses Box Content to support file uploads, and I'm handling `multipart/form-data` requests using `busboy`, but when I hand off the file stream created by `busboy` to `boxClient.files.uploadFile(...)`, the latter errors out with a 500 status code.

As another data point, when I test it using a file read through `fs.createReadStream` it works fine.  However, if I pipe that stream through a PassThrough (`fs.createReadStream(...).pipe(new PassThrough())`) then it displays the above error behavior.  Creating a Readable stream which merely reads out the contents of a Buffer also fails similarly.

As an aside, I looked at `api-rquest.js` and saw this project is using `request`, which has given me erroneous behavior like above when trying to post files using streams other than those created by `fs.createReadStream`, and also erroring out when piping `fs.createReadStream` through a `PassThrough`.  Using a different library such as `got` did not have this issue in my testing.

Currently, I am getting around this by buffering files to disk, but doing even this in an efficient manner is impossible as I have to write the entire file to disk so that I can use `fs.createReadStream` to pass it off to the sdk.

I'm working on a monkeypatch of `api-request.js` as a quick test case to see if replacing `request` resolves this stream issue.  Hopefully I will be able to report back on that soon.

**Edit:** As an update, I'm going to note that efforts to make other libraries work across corporate proxies is making me consider strategies other than replacing `request`.  There're reasons `request` is so widely used, and some amount of proxy support seems to be one of them.

joedski:
After digging Request's and FormData's codebases, I'm going to close this.  I'll leave what I found for anyone else who runs into this.

The underlying issue is this:
- Request does not have any logic to do `transfer-encoding: chunked`.
- Request always tries to set a `content-length` header.
- Request asks FormData what the length of its data is so it can send the headers before sending the body.
- FormData can only provide the length of anything which it can determine the length of.  It has provisions in its code for handling ReadStreams created by `fs.createReadStream`, HTTP `IncomingMessage`s, and `request` streams.
- Other streams must specify their known length in the `knownLength` option, passed on the object for the third parameter to `formData.append()`.
- This is impossible to do with the Box SDK (and most anything else which uses Require and thus FormData under the hood)

My workaround was to do this:
- to whatever readable stream I'm passing to the Box Client, I add these properties:
  - `myStream.httpVersion = '1.0'` (Actual value is irrelevant, but might as well use a valid one.)
[...]

PR #519, opened 2020-06-23
Add support for all streams for uploading files

sujaygarlanka:
Addresses issue #463. Streams other than fs streams could not be used to upload files because these streams don't have the stream length as part of the stream, which is required to upload a file. We now allow a user to pass in the stream length through the options parameter for `Files.uploadFile` and `Files.uploadNewFileVersion`, so other streams are now supported. Solution was inspired by #143.


lib/managers/files.js
```diff
@@ -601,6 +602,7 @@ Files.prototype.promoteVersion = function(fileID, versionID, callback) {
  * @param {Object} [options] - Optional parameters
  * @param {string} [options.content_created_at] - RFC 3339 timestamp when the file was created
  * @param {string} [options.content_modified_at] - RFC 3339 timestamp when the file was last modified
+ * @param {int} [options.content_length] - The length of the content. This is needed if not using a fs stream or string.
```
PJSimon, line 605:
```suggestion
 * @param {int} [options.content_length] - Optional length of the content. Required if content is a read stream of any type other than fs stream.
```

lib/managers/files.js
```diff
@@ -613,10 +615,16 @@ Files.prototype.uploadFile = function(parentFolderID, filename, content, options
 		options = {};
 	}
 
+	var formOptions = {};
+	if (options && options.hasOwnProperty('content_length')) {
+		formOptions.knownLength = options.content_length;
+		delete options.content_length;
```
PJSimon, line 621: What does this do?

lib/managers/files.js
```diff
@@ -635,6 +643,7 @@ Files.prototype.uploadFile = function(parentFolderID, filename, content, options
  * @param {Object} [options] - Optional parameters
  * @param {string} [options.content_modified_at] - RFC 3339 timestamp when the file was last modified
  * @param {string} [options.name] - A new name for the file
+ * @param {int} [options.content_length] - The length of the content. This is needed if not using a fs stream or string.
```
PJSimon, line 646:
```suggestion
 * @param {int} [options.content_length] - Optional length of the content. Required if content is a read stream of any type other than fs stream.
```

lib/managers/files.js
sujaygarlanka: This gets the content_length and sets it to the key `knownLength` so the request library understands this is the length. Then it deletes it from options that are passed to create headers, body, etc.

PJSimon:
```suggestion
		// Delete content_length from options so it's not added to the request object later
		delete options.content_length;
```

PJSimon: @sujaygarlanka I think that makes sense.  I added a comment to reflect that.  Did I capture it correctly?

lib/managers/files.js
```diff
@@ -650,11 +660,17 @@ Files.prototype.uploadNewFileVersion = function(fileID, content, options, callba
 	var apiPath = urlPath(BASE_PATH, fileID, '/content'),
 		multipartFormData = {};
 
+
+	var formOptions = {};
 	if (options) {
+		if (options.hasOwnProperty('content_length')) {
+			formOptions.knownLength = options.content_length;
+			delete options.content_length;
```
sujaygarlanka, line 668:
```suggestion
			// Delete content_length from options so it's not added to the attributes of the form
			delete options.content_length;
```

salazarm:
It still doesn't work with `Readable.from(['mystringhere'])` ? =(

PR #524, opened 2020-07-21
Update metadata query documentation

sujaygarlanka:
Current implementation already supports the `fields` parameter for metadata query. Just updated the documentation to reflect this.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -907,7 +907,7 @@ client.metadata.deleteCascadePolicy(policyID)
 Query
 ---------------------
 
-To query box items based on their metadata, call [`metadata.query(from, ancestorFolderId, options, callback)`][query] with the metadata template and the folder ID to restrain the query. Additional options like the a specific query, a marker, etc. can be passed in through the options object.
+To query box items based on their metadata, call [`metadata.query(from, ancestorFolderId, options, callback)`][query] with the metadata template and the folder ID to restrain the query. Additional options like the a specific query, a marker, etc. can be passed in through the options object. **Important**: There are two types of response based on whether the fields parameter is passed into the options parameter or not. The example directly below shows the response if the fields parameter is not passed in.
 
 <!-- sample post_metadata_queries_execute_read -->
 ```js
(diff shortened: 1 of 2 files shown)


docs/metadata.md
```diff
@@ -907,7 +907,7 @@ client.metadata.deleteCascadePolicy(policyID)
 Query
 ---------------------
 
-To query box items based on their metadata, call [`metadata.query(from, ancestorFolderId, options, callback)`][query] with the metadata template and the folder ID to restrain the query. Additional options like the a specific query, a marker, etc. can be passed in through the options object.
+To query box items based on their metadata, call [`metadata.query(from, ancestorFolderId, options, callback)`][query] with the metadata template and the folder ID to restrain the query. Additional options like the a specific query, a marker, etc. can be passed in through the options object. **Important**: There are two types of response based on whether the fields parameter is passed into the options parameter or not. The example directly below shows the response if the fields parameter is not passed in.
```
PJSimon, line 910:
```suggestion
To query Box items based on their metadata, call [`metadata.query(from, ancestorFolderId, options, callback)`][query] with the metadata template and the folder ID to restrain the query. Additional options like a specific query, a marker, etc. can be passed in through the options object. **Important**: One of two possible response types are returned, based on whether the fields parameter is passed into the options parameter or not. The example directly below shows the response if the fields parameter is not passed in.
```

docs/metadata.md
```diff
@@ -932,110 +932,166 @@ client.metadata.query(from, ancestorFolderId, options)
...
+								"$version": 0,
+								"customerName": "Phoenix Corp",
+								"$type": "someTemplate-3d5fcaca-f496-4bb6-9046-d25c37bc5594",
+								"$typeVersion": 0,
+								"$id": "ba52e2cc-371d-4659-8d53-50f1ac642e35",
+								"amount": 100,
+								"claimDate": "2016-04-10T00:00:00Z",
+								"region": "West",
+								"$typeScope": "enterprise_123456"
+							}
```
PJSimon, line 1030: Hard to tell if anything changed other than the indent here.

sujaygarlanka: Just reformatting. No data changed

PR #538, opened 2020-08-25
Fix incorrect headers in Zip documentation

cbetta:
Headers we're not using correct markdown syntax.


cbetta:
Superceded by #539

PR #547, opened 2020-10-02
Add support for search param to get shared link items

sujaygarlanka:
Did manual testing.


lib/managers/search.js
```diff
@@ -76,6 +76,8 @@ Search.prototype = {
 	 * @param {string} [options.type] - The type of objects you want to include in the search results. The type can be file, folder, or web_link
 	 * @param {string} [options.trash_content=non_trashed_only] - Controls whether to search in the trash. The value can be trashed_only or non_trashed_only
 	 * @param {SearchMetadataFilter[]} [options.mdfilters] - Searches for objects with a specific metadata object association.  Searches with the this parameter do not require a query string
+	 * @param {boolean} [options.are_shared_links_included] - Determines whether to include items accessible only via shared link in the response. To include shared link items, a user must also pass in 'is_only_accessible_via_shared_link' to the fields parameter
+	 * @param {string} [options.fields] - Comma-delimited list of fields to be included in the response
```
PJSimon, line 80:
```suggestion
	 * @param {boolean} [options.include_recent_shared_links] - Determines whether to include items accessible only via shared link in the response.
```

sujaygarlanka: Will make the change. Should keep fields though because it is a parameter that is part of the api but does not exist for this method.

PJSimon (approved):

Issue #533, opened 2020-08-19
Passing the body as an Object to validateWebhookMessage is not reliable


Avaq:
- [x] I have checked that the [SDK documentation][sdk-docs] doesn't solve my issue.
- [x] I have checked that the [API documentation][api-docs] doesn't solve my issue.
- [x] I have searched the [Box Developer Forums][dev-forums] and my issue isn't already reported (or if it has been reported, I have attached a link to it, for reference).
- [x] I have searched [Issues in this repo][github-repo] and my issue isn't already reported.

### Description of the Issue

It was discovered via a recent Box Customer Success support ticket (`#2204084`) that using the object passing style implemented in #253 with `validateWebhookMessage` is not reliable.

The problem is that `JSON.parse` is lossy. The code added in #253\* assumes that `JSON.stringify(JSON.parse(x)) === x`, which is not always the case. For example:

```js
const x = '"\\/"';
JSON.stringify(JSON.parse(x)) !== x
```

This exact problem occurs when a webhook sends a message regarding an updated file. These messages include escaped characters, such as the one assigned to `x` in my example, and as such, are invalidated by `validateWebhookMessage` after having been JSON parsed.

\* The problematic code:

https://github.com/box/box-node-sdk/blob/f91ddc468aade7067c0aafa3ec4eb6944f2e911e/lib/managers/webhooks.js#L351-L355

### Versions Used
Node SDK: 1.33.0

[sdk-docs]: ./doc
[api-docs]: https://developer.box.com/docs
[dev-forums]: https://community.box.com/t5/Platform-and-Development-Forum/bd-p/DeveloperForum
[github-repo]: https://github.com/box/box-java-sdk/search?type=Issues

PJSimon:
Hi @Avaq, 

Thanks for reporting this issue.  We are working on a fix now!

Thanks,

Patrick

PR #568, opened 2021-02-06
Fix webhook signature validation

swfree:
### Issue Link :link:

- Fixes https://github.com/box/box-node-sdk/issues/533

### Implementation Details :construction:

- PHP's json_encode method escapes slashes & multibyte Unicode characters by default, whereas Node's JSON.parse and JSON.stringify methods do not. This change translates the parsed JSON body back into a string that will correctly match the original body used to compute the webhook signature.

### DIFF
--- a/lib/managers/webhooks.js
+++ b/lib/managers/webhooks.js
@@ -110,6 +110,17 @@ function validateDeliveryTimestamp(headers, maxMessageAge) {
 	return true;
 }
 
+/**
+ * Stringify JSON with escaped multibyte Unicode characters to ensure computed signatures match PHP's default behavior
+ *
+ * @param {Object} body - The parsed JSON object
+ * @returns {string} - Stringified JSON with escaped multibyte Unicode characters
+ * @private
+ */
+function jsonStringifyWithEscapedUnicode(body) {
+	return JSON.stringify(body).replace(/[\u007f-\uffff]/g, char => `\\u${`0000${char.charCodeAt(0).toString(16)}`.slice(-4)}`);
+}
+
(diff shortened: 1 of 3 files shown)


lib/managers/webhooks.js
```diff
@@ -110,6 +110,17 @@ function validateDeliveryTimestamp(headers, maxMessageAge) {
...
 
+/**
+ * Stringify JSON with escaped multibyte Unicode characters to ensure computed signatures match PHP's default behavior
+ *
+ * @param {Object} body - The parsed JSON object
+ * @returns {string} - Stringified JSON with escaped multibyte Unicode characters
+ * @private
+ */
+function jsonStringifyWithEscapedUnicode(body) {
+	return JSON.stringify(body).replace(/[\u007f-\uffff]/g, char => `\\u${`0000${char.charCodeAt(0).toString(16)}`.slice(-4)}`);
```
sujaygarlanka, line 121: Whats the logic here?

swfree: This code snippet escapes all non-ASCII characters in their unicode escape sequence "\\u....". `charCodeAt()` returns the UTF-16 code unit from 0-65535, and `toString(16)` returns the hexademical value. So "007f"-"ffff" is numbers 127-65535, which is how the regex captures all the non-ASCII characters.

This is needed to match PHP's `json_encode` which also encodes all non-ASCII characters using "\u...." escape sequences. This stack overflow answer has a good explanation: https://stackoverflow.com/a/22745663/12092405

sujaygarlanka (dismissed):

sujaygarlanka (approved):

PR #582, opened 2021-04-19
Intro TypeScript migration

### LINKED ISSUE #170: Create Typescript definitions
My organization is using the Box Node SDK in a Typescript project, and we've been creating our own type definitions for all the pieces of the SDK that we use, but would benefit from a types library contributed to by Box (and any other contributors of the SDK) so we don't have to build types every time we start using a new feature of the Box API.

Are there any plans for Box to build some Typescript definitions for this SDK?
If not, I can submit a PR to either this repo or DefinitelyTyped with some of the types I've already created. Any preference there?

mhagmajer:
This is a first step in our migration to TypeScript.

- All JS source code files have been moved to `/src` which will eventually host the TypeScript code
- I added another step to Makefile which transpiles files from `/src` to `/lib`. Since almost all files in `/src` are JS files it just copies them over
- The compiled `/lib` directory includes source maps and TypeScript declarations for the original files
- `box-client.js` has already been moved to `.ts` with minimal changes due to some issues
- all tests are run against the transpiled code inside `/lib` directory

Notes:
- `/lib` was added to `.gitignore` since it now includes generated code. `/lib` directory should still be included in the package

Closes #170.


package.json
```diff
@@ -52,6 +52,7 @@
     "eslint-plugin-unicorn": "^4.0.3",
     "istanbul": "^0.4.3",
     "jsdoc": "^3.6.3",
+    "jsonlint": "^1.6.3",
```
mhagmajer, line 55: Running `node Makefile.js` failed both on my machine and Travis without this dependency

Makefile.js
```diff
@@ -70,6 +70,9 @@ target.lint = function() {
 
 	var code = 0;
 
+	echo('Compiling TypeScript code');
+	code += nodeCLI.exec('./node_modules/.bin/tsc').code;
```
swfree, line 74: What do you think about moving this to a new `target.build` function specifically for this, and then add a "build" script to package.json that just runs `node Makefiles.js build`? I'm thinking of our release process in particular 🤔

.gitignore
```diff
@@ -10,3 +10,5 @@ docs/jsdoc-dev
 .nyc_output
 .vscode/
 .npmrc
+
+lib/
```
swfree, line 14: minor: missing newline

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "skipLibCheck": true,
+    "allowJs": true,
+    "sourceMap": true,
+    "declaration": true,    
+    "noImplicitAny": false,
+    "noImplicitThis": false,
+    "forceConsistentCasingInFileNames": true
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules", "**/*.spec.ts", "src/box-client.js"]
```
swfree, line 17: Is `src/box-client.js` needed anymore, since its been renamed to `src/box-client.ts`?

Makefile.js
mhagmajer: Good idea, applied the changes!

swfree (approved):

Makefile.js
```diff
@@ -85,6 +95,7 @@ target.lint = function() {
 
 target.test = function() {
 	var code = target.lint();
+	code += target.build();
```
swfree, line 98: The travis builds should pass if you move target.build() before target.lint()

swfree (approved):

PR #584, opened 2021-04-20
Add type annotations to box-client.ts

mhagmajer:
This moves `BoxClient` into a modern class syntax and adds type annotations described in comments into TypeScript space. We also drop `noImplicitAny`, `noImplicitThis` and `skipLibCheck` from `tsconfig` for stricter type checking.

Note: some values are still marked as `any`. We'll figure out specific types for them as we move further in this migration.

This also uses Prettier for formatting of the box-client.ts source code.

### DIFF
--- a/src/box-client.ts
+++ b/src/box-client.ts
@@ -18,13 +18,13 @@ type CollaborationRole = string;
  * A Box file or folder type constant
  * @typedef {string} ItemType
  */
-type ItemType = string;
+type ItemType = 'file' | 'folder';
 
 /**
  * An access level constant. Used for setting and updating shared links, folder upload, etc.
  * @typedef {?Object} AccessLevel
  */
-type AccessLevel = Object | null;
+type AccessLevel = object | null /* FIXME */;
 
 type APISession = any /* FIXME */;
 type APIRequestManager = any /* FIXME */;
@@ -141,8 +141,8 @@ function getFullURL(defaultBasePath: string, url: string) {
  * @returns {Object} The batch API request object
  * @private
  */
-function formatRequestForBatch(params: Record<string, string>) {
-	var relativePath = params.url.replace(/^http.*?\/\d\.\d\//, '/');
+function formatRequestForBatch(params: Record<string, string | object>) {
+	var relativePath = (params.url as string).replace(/^http.*?\/\d\.\d\//, '/');
 
 	return {
 		method: params.method,
@@ -306,7 +306,7 @@ class BoxClient {
 	 * @returns {Object} - a new object with the headers needed for the request
 	 * @private
 	 */
-	_createHeadersForRequest(callerHeaders: Object | null, accessToken: string) {
+	_createHeadersForRequest(callerHeaders: object | null, accessToken: string) {
 		var headers: Record<string, string> = {};
 
 		// 'Authorization' - contains your valid access token for authorization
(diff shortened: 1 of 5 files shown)


src/box-client.ts
```diff
@@ -141,15 +141,15 @@ function getFullURL(defaultBasePath, url) {
  * @returns {Object} The batch API request object
  * @private
  */
-function formatRequestForBatch(params) {
-
+function formatRequestForBatch(params: Record<string, string>) {
```
swfree, line 144: Is `params` always shaped like `string: string`? Is `params.body` not `string: object` in this case?

src/box-client.ts
```diff
@@ -13,19 +12,22 @@
...
+ * A Box file or folder type constant
+ * @typedef {string} ItemType
+ */
+type ItemType = string;
 
 /**
  * An access level constant. Used for setting and updating shared links, folder upload, etc.
  * @typedef {?Object} AccessLevel
  */
+type AccessLevel = Object | null;
```
swfree, line 27: Should we add a `/* FIXME */` note to update this type to be more specific later on? Or will it remain defined as `Object`?

src/box-client.ts
```diff
@@ -13,19 +12,22 @@
...
+type CollaborationRole = string;
 
 /**
-* A Box file or folder type constant
-* @typedef {string} ItemType
-*/
+ * A Box file or folder type constant
+ * @typedef {string} ItemType
+ */
+type ItemType = string;
```
swfree, line 21: What are your thoughts on making this more specific, like `type ItemType = 'file' | 'folder'`?

src/box-client.ts
swfree: Also - still learning more the nuances of typescript myself, but it looks like we want lowercase `object` here instead of uppercase `Object`?

- https://www.typescriptlang.org/docs/handbook/basic-types.html#about-number-string-boolean-symbol-and-object
- https://2ality.com/2020/01/typing-objects-typescript.html#object-vs.-object%3A-primitive-values

src/box-client.ts
```diff
@@ -174,182 +173,643 @@ function formatResponseForBatch(response) {
...
+	 * @param {Object} [options] - Optional parameters
+	 * @param {ActorParams} [options.actor] - Optional actor parameters for creating annotator tokens with Token Auth client
+	 * @param {SharedLinkParams} [options.sharedLink] - Optional shared link parameters for creating tokens using shared links
+	 * @param {Function} [callback] Called with the new token
+	 * @returns {Promise<TokenInfo>} A promise resolving to the exchanged token info
+	 */
+	exchangeToken(
+		scopes: string | string[],
+		resource: string,
+		options: Object,
```
swfree, line 504: Looks like `options` can sometimes be a function as well, looking at line 508 below

src/box-client.ts
```diff
@@ -174,182 +173,643 @@ function formatResponseForBatch(response) {
...
+			body: {
+				requests: this._batch.map((batchReq: any /* FIXME */) =>
+					formatRequestForBatch(batchReq.params)
+				),
+			},
+		};
+
+		var batch: any[] = this._batch;
+		this._batch = null;
+		return this.post('/batch', params, /* callback */ undefined)
```
swfree, line 665: Is the addition of `/* callback */ undefined` needed?

src/box-client.ts
mhagmajer: yes, good point

src/box-client.ts
mhagmajer: good catch, this is actually an optional argument

mhagmajer:
Thanks @swfree, I've applied your comments!

swfree (approved):

PR #586, opened 2021-04-26
ts migrate: box-node-sdk.ts

mhagmajer:
This migrates box-node-sdk.js to TypeScript and thus provides more type information to the users of the SDK. The following step will include going deeper into the stack and migrating more files


src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+var Config = require('./util/config'),
+	BasicAPISession = require('./sessions/basic-session'),
+	PersistentAPISession = require('./sessions/persistent-session'),
+	AppAuthSession = require('./sessions/app-auth-session'),
+	Webhooks = require('./managers/webhooks');
```
swfree, line 57: Should we move these requirements to be grouped with the imports on lines 5-10? We could move the "Requirements" header up there too

src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+ * 1. Provides getters to spawn client instances for users to interact with the Box API.
+ * 2. Provides manual capability to acquire tokens via token grant endpoints.
+ *    However, it is recommended to use clients to do this for you.
+ * 3. Emits notification events about relevant request/response events. Useful for logging Box API interactions.
+ *    Notification events: request retries, exceeding max retries, permanent failures.
+ *
+ * @param {UserConfigurationOptions} params User settings used to initialize and customize the SDK
+ * @constructor
+ */
+class BoxSDKNode extends EventEmitter {
```
swfree, line 80: Nice improvement 👍

src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+		this.config = new Config(params);
+		this._eventBus = eventBus;
+		this._setup();
+	}
+
+	/**
+	 * Setup the SDK instance by instantiating necessary objects with current
+	 * configuration values.
+	 *
+	 * @param {EventEmitter} eventBus The event bus to use
```
swfree, line 129: I see that this is carried over from the original file, so if this is a bug it's probably an old bug 🤔   But it looks like `_setup` doesn't take any params. Should we remove this line?

src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+			params.clientSecret = boxAppSettings.clientSecret;
+		}
+
+		// Only try to assign app auth settings if they are present
+		// Some configurations do not include them (but might include other info, e.g. webhooks)
+		if (
+			typeof boxAppSettings.appAuth === 'object' &&
+			boxAppSettings.appAuth.publicKeyID
+		) {
+			var appAuth: Exclude<typeof params['appAuth'], void> = {};
```
swfree, line 189: Slight preference for `params.appAuth` syntax, to be consistent with the rest of the file

src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+	 * managing App Users (as the enterprise) or performing operations as the App
+	 * Users themselves (as a user).
+	 *
+	 * @param {string} type The type of entity to operate as, "enterprise" or "user"
+	 * @param {string} id The Box ID of the entity to operate as
+	 * @param {TokenStore} [tokenStore] (Optional) the token store to use for caching tokens
+	 * @returns {BoxClient} A new client authorized as the app user or enterprise
+	 */
+	getAppAuthClient(type: string, id: string, tokenStore?: TokenStore) {
+		if (type === 'enterprise' && !id) {
```
swfree, line 299: Based on this conditional, it looks like `id` is an optional param (even though the comment docs don't indicate it is optional)

src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+	 * Refreshes the access and refresh tokens for a given refresh token.
+	 *
+	 * @param {string} refreshToken - A valid OAuth refresh token
+	 * @param {TokenRequestOptions} [options] - Sets optional behavior for the token grant, null for default behavior
+	 * @param {Function} [callback] - passed a TokenInfo object if tokens were granted successfully
+	 * @returns {Promise<TokenInfo>} Promise resolving to the token info
+	 */
+	getTokensRefreshGrant(
+		refreshToken: string,
+		options: TokenRequestOptions | null,
```
swfree, line 363: should this be `options: TokenRequestOptions | Function | null,` based on line 366

src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+	/**
+	 * Gets tokens for App Users via a JWT grant
+	 * @param {string} userID The ID of the App User to generate a token for
+	 * @param {TokenRequestOptions} [options] - Sets optional behavior for the token grant, null for default behavior
+	 * @param {Function} [callback] Passed the tokens if successful
+	 * @returns {Promise<TokentInfo>} Promise reolving to the token infp
+	 */
+	getAppUserTokens(
+		userID: string,
+		options: TokenRequestOptions | null,
```
swfree, line 415: same here - add `Function`?

src/box-node-sdk.ts
```diff
@@ -0,0 +1,463 @@
...
+	 * Revokes a token pair associated with a given access or refresh token.
+	 *
+	 * @param {string} token - A valid access or refresh token to revoke
+	 * @param {TokenRequestOptions} [options] - Sets optional behavior for the token grant, null for default behavior
+	 * @param {Function} [callback] - If err, revoke failed. Otherwise, revoke succeeded.
+	 * @returns {Promise<TokenInfo>} Promise resolving to the token info
+	 */
+	revokeTokens(
+		token: string,
+		options: TokenRequestOptions | null,
```
swfree, line 438: same here - add `Function`?

swfree (approved):

PR #590, opened 2021-04-26
ts migrate: src/util/errors.ts

mhagmajer:


src/util/errors.ts
```diff
@@ -76,7 +82,7 @@ module.exports = {
 	 * @param {string} message - the response error message
 	 * @returns {Errors~ResponseError} an error describing the response error
 	 */
-	buildResponseError(response, message) {
+	buildResponseError(response: any /* FIXME */, message: string) {
```
swfree, line 85: Looks like `message` is optional (based on line 87)

src/util/errors.ts
```diff
@@ -133,8 +140,7 @@ module.exports = {
 	 * @param {string} [message] - Optional message for the error
 	 * @returns {Errors~AuthError} A properly formatted authentication error
 	 */
-	buildAuthError(response, message) {
-
+	buildAuthError(response: any /* FIXME */, message: string) {
```
swfree, line 143: looks like `message` is optional based on the line below

swfree (approved):

PR #593, opened 2021-04-27
ts migrate: src/sessions/persistent-session.ts

mhagmajer:


swfree:
Build failures are coming from typescript compilation errors: `Property 'unwrapAndThrow' does not exist on type 'typeof import("/home/travis/build/box/box-node-sdk/src/util/errors")'` -- https://travis-ci.com/github/box/box-node-sdk/jobs/501826702

Should be fixed if you update this to `export =` https://github.com/box/box-node-sdk/blob/main/src/util/errors.ts#L76

src/sessions/persistent-session.ts
```diff
@@ -0,0 +1,307 @@
...
+ */
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+import { Promise } from 'bluebird';
+import assert = require('assert');
+import errors = require('../util/errors');
+import httpStatusCodes = require('http-status');
```
swfree, line 12: These imports all work the same if they're formatted like `import assert from 'assert';`, right? Wondering if we should settle on that format for consistency

mhagmajer: Yes, once all dependencies are translated to TypeScript (and using `export` syntax) it's okay we can stick to `import ... from` syntax. I've made updates to the PR. Thanks!

mhagmajer:
Thanks for the review @swfree! I've applied the changes you requested.

swfree (approved):

PR #598, opened 2021-04-28
ts migrate: src/api-request.ts

mhagmajer:


src/api-request.ts
```diff
@@ -0,0 +1,398 @@
...
+
+// ------------------------------------------------------------------------------
+// Private
+// ------------------------------------------------------------------------------
+
+// Message to replace removed headers with in the request
+var REMOVED_HEADER_MESSAGE = '[REMOVED BY SDK]';
+
+// Range of SERVER ERROR http status codes
+var HTTP_STATUS_CODE_SERVER_ERROR_BLOCK_RANGE = [500, 599];
```
swfree, line 106: should we update these to `const`?

swfree (approved):

src/api-request.ts
mhagmajer: Will do in a follow up

PR #601, opened 2021-04-29
ts migrate: src/event-stream.ts

mhagmajer:


src/event-stream.ts
```diff
@@ -0,0 +1,362 @@
...
+ */
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+import { Promise } from 'bluebird';
+import qs from 'querystring';
+import { Readable } from 'stream';
+import util from 'util';
```
swfree, line 12: Looks like `util` is no longer used in this file

swfree (approved):

src/event-stream.ts
mhagmajer: Thanks! Will fix in a follow-up.

PR #611, opened 2021-05-04
ts migrate: src/managers/legal-hold-policies.ts

mhagmajer:


src/managers/legal-hold-policies.ts
```diff
@@ -0,0 +1,391 @@
...
+	 * @param {string} [updates.release_notes] - Notes around why the policy was released
+	 * @param {Function} [callback] - Passed the updated policy information if it was acquired successfully, error otherwise
+	 * @returns {Promise<Object>} A promise resolving to the updated policy
+	 */
+	update(
+		policyID: string,
+		updates: {
+			policy_name: string;
+			description: string;
+			release_notes: string;
```
swfree, line 137: based on the API, these should probably be optional types https://developer.box.com/reference/put-legal-hold-policies-id/#body-parameters

swfree (dismissed):

src/managers/legal-hold-policies.ts
mhagmajer: right, thanks! :)

PR #625, opened 2021-05-10
Insensitive language: replace whitelist with allowlist

mhagmajer:
This replaces uses of `*whitelist*` with `*allowlist*` across file names and contents.

Some of the code is related to the actually endpoint include the insensitive word:
- https://developer.box.com/reference/get-collaboration-whitelist-entries/
- https://developer.box.com/reference/get-collaboration-whitelist-exempt-targets/

Note for reviewer:

This PR includes whitespace changes (trailing newlines removed). It's useful to turn on this setting when reviewing:
Screenshot 2021-05-10 at 18 02 32


sujaygarlanka:
We probably want to keep collaboration_whitelist.ts alongside collaboration_allowlist.ts, but just deprecate it. This will allow customers to take advantage of updates to our SDK and not make this a breaking change. May make sense to keep the documentation and tests along with it, but deprecate them as well. I did something like this here https://github.com/box/box-java-sdk/pull/889. Thoughts?

CHANGELOG.md
```diff
@@ -211,7 +211,7 @@ to accept the request body as an `Object`
 
 - Added options to preserve file timestamps on [file upload](./docs/files.md#upload-a-file)
 and to rename a file or preserve modification timestamp on [new version upload](./docs/files.md#upload-a-new-version-of-a-file)
-- Added [Collaboration Whitelist](./docs/collaboration-whitelist.md) functionality to allow enterprise admins
```
sujaygarlanka, line 214: Shouldn't change this because it is a history of changes.

docs/README.md
```diff
@@ -9,7 +9,7 @@ the SDK are available by topic:
 * [Client](client.md)
 * [Classifications](classifications.md)
 * [Collaborations](collaborations.md)
-* [Collaboration Whitelist](collaboration-whitelist.md)
```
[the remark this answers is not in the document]
mhagmajer: Thanks @sujaygarlanka. Maybe instead of doing the text replace changes to the docs folder I would just add notes to specific classes and functions that were renamed, thoughts? Something to note is that the docs are automatically generated from the code and this process will need to be updated anyway once we migrated full to TypeScript.

CHANGELOG.md
sujaygarlanka: We should add to the CHANGELOG.md, but probably not update older history because it would be inaccurate and the commits associated with the change will no longer match the change.

src/box-client.ts
```diff
@@ -288,10 +288,13 @@ class BoxClient {
 		this.devicePins = new DevicePins(this);
 		this.webhooks = new Webhooks(this);
 		this.recentItems = new RecentItems(this);
-		this.collaborationWhitelist = new CollaborationWhitelist(this);
+		this.collaborationAllowlist = new CollaborationAllowlist(this);
 		this.termsOfService = new TermsOfService(this);
 		this.storagePolicies = new StoragePolicies(this);
 
+		// Legacy insensitive language
```
sujaygarlanka, line 295: Will this work for methods like `this.collaborationWhitelist.getAllWhitelistedDomains`? It seems like it will be `this.collaborationWhitelist.getAllAllowlistedDomains`, which means there still may be a breaking change.

docs/README.md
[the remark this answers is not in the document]
sujaygarlanka: @mhagmajer That works as well. I would also see if there is a good way to communicate in the code the all whitelist methods are deprecated, like adding a deprecated decorator or something of that sort to `whitelist` methods.

src/box-client.ts
mhagmajer: @sujaygarlanka I think it will because `CollaborationWhitelist` now includes these additional aliases:

```js
// Legacy insensitive language
(CollaborationAllowlist.prototype as any).getWhitelistedDomain =
	CollaborationAllowlist.prototype.getAllowlistedDomain;
(CollaborationAllowlist.prototype as any).getAllWhitelistedDomains =
	CollaborationAllowlist.prototype.getAllAllowlistedDomains;
```

CHANGELOG.md
mhagmajer: Ok, I'm reverting these

docs/README.md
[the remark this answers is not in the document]
mhagmajer: Good idea, I added the `@deprecated` annotations to:
- `collaborationWhitelist` on `BoxClient`
- `getWhitelistedDomain` and `getAllWhitelistedDomains` on `CollaborationAllowlist`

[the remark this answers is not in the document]
mhagmajer: @sujaygarlanka I updated  `docs/collaboration-whitelist.md` but left jsdocs as is - these are generated docs and we should regenerate them from TypeScript in a separate PR

sujaygarlanka (dismissed):

sujaygarlanka (approved):

PR #626, opened 2021-05-17
Remove insensitive language CSS class from docs

mhagmajer:


sujaygarlanka (commented):
Won't this be added once docs are regenerated?

mhagmajer:
> Won't this be added once docs are regenerated?

It would, however, the doc generation process will need to be updated anyway now that we have sources in TypeScript. That's why I decided to just remove the classes from generated code for now

sujaygarlanka (approved):

PR #648, opened 2021-07-29
Add sensitive event types for admin invites

mhagmajer:


src/managers/events.ts
```diff
@@ -71,6 +73,8 @@ enum EventType {
 	ITEM_UNSYNC = 'ITEM_UNSYNC',
 	LOCK = 'LOCK',
 	LOGIN = 'LOGIN',
+	MASTER_INVITE_ACCEPT = 'MASTER_INVITE_ACCEPT',
+	MASTER_INVITE_REJECT = 'MASTER_INVITE_REJECT',
```
mwwoda, line 77: We don't need that as we have `ADMIN_INVITE_ACCEPT` and `ADMIN_INVITE_REJECT`.

mwwoda (approved):

PR #673, opened 2021-10-02
docs: link to CHANGELOG.md (for npm)

coolaj86:
There's no clear way to view the CHANGELOG from npm, which is where most searches will lead.

Since you just bumped to v2.0.0, I thought it might help others like myself to get to the necessary information faster.

(and for those finding this comment via a search for `2.0.0`, it's a small breaking change - removing node v6.0 support to fix a security vuln)


antusus (approved):
@coolaj86 Thank you for that. It is a good idea to link the change log into the readme file.

PR #681, opened 2021-12-02
docs: Added example of using uploadFile with Readable.

antusus:


docs/files.md
```diff
@@ -303,21 +303,40 @@ client.files.uploadFile(folderID, 'My File.pdf', stream)
...
 ```
 
+If you want to pass a Readable you must pass the content length as an optional parameter.
+Here is an example of passing Base64 String as file content:
+```js
+var {Readable} = require('stream');
+
+var base64Content = 'TXkgY29udGVudAo='; // your base64 content
+var base64Buffer = Buffer.from(base64Content, 'base64');
+// we are using just Readable to creata stream, but you can use any library you want
```
mhagmajer, line 327:
```suggestion
// we are using just Readable to create a stream, but you can use any library you want
```

mhagmajer (dismissed):
👌

PR #685, opened 2022-01-17
fix: add new SharedLinkAccessLevel enum with correct type

mwwoda:
SDK is incorrectly sending access when creating/updating SharedLinks - `{\"access\":{\"access\":\"open\"}` instead of `{\"access\": \"open\"}`


mwwoda:
PR is no longer needed

PR #686, opened 2022-02-03
feat: Deprecating `index_name` when executing metadata query

arjankowski:


src/managers/metadata.ts
```diff
@@ -413,6 +415,8 @@ class Metadata {
 			ancestor_folder_id: ancestorFolderId,
 		};
 
+		delete options?.index_name;
```
mhagmajer, line 418: I don't think it's necessary to delete this property from the options object

mhagmajer:
```ts
const { index_name, ...newOptions } = options
```

mhagmajer (approved):

PR #709, opened 2022-03-15
feat: Client Credentials Grant authentication method

antusus:


antusus:
May main concern is using AnonymousSession with CCG when logging as application or user. Maybe those sessions should be different? I'm creating new instances of those sessions when you get client for Application or User account so there is a separation. But is it good enough?

mhagmajer (approved):
🙌🏻

PR #744, opened 2022-07-15
feat: Added support update and remove user avatar

congminh1254:
Closes: SDK-2187


mwwoda (approved):
LGTM! I guess you didn't have problems with the mimetype like in other SDKs.

PR #750, opened 2022-08-04
fix: Fix error while generate token using OAuth

congminh1254:
I am fixing issue [#286](https://github.com/box/boxcli/issues/286) of boxcli.

Error happen when trying to generate token while using OAuth auth, I fixed to it show the correct error message:
```bash
> box tokens:get -u 20111795585
Must provide app auth configuration to use JWT Grant
```


lukaszsocha2 (approved):
Please when merging in commit body add: 
Fixes: #286
it will close issue automatically after merging

PR #759, opened 2022-09-13
feat: Added support for Access Only Collaboration 

antusus:


package-lock.json
```diff
@@ -2817,9 +2817,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30001252",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001252.tgz",
-      "integrity": "sha512-I56jhWDGMtdILQORdusxBOH+Nl/KgQSdDmpJezYddnAkVOmnoU8zwjTV9xAjMIYxr0iPreEAVylCGcmHCjfaOw==",
+      "version": "1.0.30001399",
```
lukaszsocha2, line 2820: do we want to commit that ?

src/schemas/collaboration.ts
```diff
@@ -0,0 +1,122 @@
...
+	 */
+	expires_at?: string;
+	/**
+	 * The email address used to invite an unregistered collaborator, if they are not a registered user.
+	 */
+	invite_email?: string;
+	/**
+	 * The file or folder to which access is granted. The field is null when the collaboration status is pending.
+	 */
+	item?: any;
```
lukaszsocha2, line 100: item?: File | Folder | WebLink ?

tests/integration_test/__tests__/collaborations.test.js
```diff
@@ -0,0 +1,84 @@
...
+		await context.user.dispose();
+		await context.collaborator.dispose();
+		context.folder = null;
+		context.user = null;
+		context.collaboratorEmail = null;
+	});
+
+	it('creates and removes collaboration', async() => {
+		let collaboration;
+		collaboration = await context.client.collaborations.create(
```
lukaszsocha2, line 28: any specific reason why this is not inlined?

lukaszsocha2: `let collaboration = await context.client.collaborations.create(`

lukaszsocha2:
Two questions left. I pushed a commit with proposed changes. Tell me if you like them.

src/schemas/collaboration.ts
antusus: We do not have those mappings :(

package-lock.json
antusus: So the npm is complaining we are using outdated database and if you fix it it bumps `caniuse` which is fine

lukaszsocha2 (approved):

PR #760, opened 2022-09-16
feat: Add sign request webhook

congminh1254:
Closes: [SDK-2344](https://jira.inside-box.net/browse/SDK-2344)

### DIFF
--- a/tests/integration_test/__tests__/sign-request.test.js
+++ b/tests/integration_test/__tests__/sign-request.test.js
@@ -86,3 +86,72 @@ test('test sign request', async() => {
 		await file.dispose();
 	}
 });
+
+test('test sign request webhook', async() => {
+	let file = await createBoxTestFile(context.client, path.join(__dirname, '../resources/blank.pdf'), 'blank.pdf', context.folder.id);
+	try {
+		const sr = await context.client.signRequests.create({
+			signers: [
+				{
+					email: 'sdk_integration_test@boxdemo.com',
+					role: 'signer',
+				}
+			],
+			source_files: [
+				{
+					type: 'file',
+					id: file.id,
+				}
+			],
+			parent_folder: {
+				id: context.folder.id,
+				type: 'folder',
+			},
+		});
+		expect(sr.id).toBeDefined();
+		expect(sr.sign_files.files).toBeDefined();
+
+		// Create a webhook for the sign request
+		let signFileId = sr.sign_files.files[0].id;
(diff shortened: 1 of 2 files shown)


tests/integration_test/__tests__/sign-request.test.js
```diff
@@ -86,3 +86,75 @@ test('test sign request', async() => {
 		await file.dispose();
 	}
 });
+
+test('test sign request webhook', async() => {
```
antusus, line 90: This should not be here. You do not need SignRequest to create and remove webhook. Maybe just create it-tests for webhooks?

src/box-client.ts
```diff
@@ -222,7 +222,7 @@ class BoxClient {
 	weblinks: any;
 	retentionPolicies: any;
 	devicePins: any;
-	webhooks: any;
+	webhooks: Webhooks;
```
antusus, line 225: If possible we should be fixing those types whenever we can

antusus (changes requested):
Move IT tests to dedicated place

tests/integration_test/__tests__/webhook.test.js
```diff
@@ -0,0 +1,60 @@
...
+	webhook = await context.client.webhooks.update(
+		webhook.id,
+		{
+			address: 'https://example.com/sign_webhook_updated',
+			triggers: [context.client.webhooks.triggerTypes.FILE.DELETED]
+		}
+	);
+	expect(webhook.id).toBeDefined();
+	expect(webhook.address).toBe('https://example.com/sign_webhook_updated');
+	expect(webhook.triggers.length).toBe(1);
```
antusus, line 48: You could check if the update changed the trigger:
`expect(webhook.triggers[0]).toBe('FILE.DELETED');`

tests/integration_test/__tests__/webhook.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+	webhook = await context.client.webhooks.get(webhook.id);
+	expect(webhook.id).toBeDefined();
```
antusus, line 51: You could test if other fields are also set:
```js
expect(webhook.address).toBe('https://example.com/sign_webhook_updated');
expect(webhook.triggers.length).toBe(1);
expect(webhook.triggers[0]).toBe('FILE.DELETED');
```
this way you also cover GET

antusus (changes requested):

antusus (approved):
👍

PR #779, opened 2022-11-16
chore: Add fields `disposition_at` field for files under retention

congminh1254:
Closes: SDK-1920.
No change needed for this task, so I just add unit tests and integration test.


tests/endpoint-test.js
```diff
@@ -1152,6 +1172,35 @@ describe('Endpoint', () => {
...
 					});
 			});
+			it('should extend the retention policy expiration date for the file', done => {
+				const fileID = '1234567890';
+				const updates = {
+					disposition_at: '2023-12-12T10:53:43-08:00',
+				};
+				const fixture = getFixture('files/put_files_disposition_at_custom_params_200');
+				apiMock
+					.put(`/2.0/files/${fileID}?fields=disposition_at`, updates)
```
antusus, line 1182: I do not think API supports PUT with query params?
https://developer.box.com/reference/put-files-id/

arjankowski (approved):
👍

antusus (changes requested):
There is some strange code in `files.ts`

```ts
if (updates && updates.fields) {
	params.qs = {
		fields: updates.fields,
	};
	delete updates.fields;
}
```
why do we put body of a put in the query? That is why you had to put this to the test.

tests/endpoint-test.js
congminh1254: Hi @antusus 

It's have `fields` query params here: https://developer.box.com/reference/put-files-id/#param-file_id

PR #797, opened 2023-01-24
fix: return empty object when `tos` status is not present

mwwoda:
return empty object when tos status is not present
handle 201 code correctly in `setUserStatus`

### DIFF
--- a/src/managers/terms-of-service.ts
+++ b/src/managers/terms-of-service.ts
@@ -274,7 +274,9 @@ class TermsOfService {
 				if (response.statusCode !== 200) {
 					throw errors.buildUnexpectedResponseError(response);
 				}
-				return response.body.entries[0];
+				return response.body.entries.length != 0
+					? response.body.entries[0]
+					: {};
 			})
 			.asCallback(callback);
 	}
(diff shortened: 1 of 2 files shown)


src/managers/terms-of-service.ts
```diff
@@ -274,7 +274,9 @@ class TermsOfService {
 				if (response.statusCode !== 200) {
 					throw errors.buildUnexpectedResponseError(response);
 				}
-				return response.body.entries[0];
+				return response.body.entries.length != 0
+					? response.body.entries[0]
+					: {};
```
mhagmajer, line 279:
```suggestion
				return response.body.entries[0] ?? {};
```

mhagmajer:
@mwwoda do you know what's up with the failing integration test?

mwwoda:
Not quite sure about the reason of the failed test. Could just be a flaky one. After rerun it passes.

src/managers/terms-of-service.ts
mwwoda: fixed

mhagmajer (approved):
👍🏻

PR #806, opened 2023-02-27
chore: Add tests for retrying strategy

congminh1254:
Verify the retrying strategy for Node SDK and added some tests.

### DIFF
--- a/tests/lib/api-request-test.js
+++ b/tests/lib/api-request-test.js
@@ -356,6 +356,7 @@ describe('APIRequest', function() {
 
 			var apiRequest = new APIRequest(config, eventBusFake);
 			apiRequest.execute(function callback() {
+				assert.equal(apiRequest.numRetries, 1);
 				done();
 			});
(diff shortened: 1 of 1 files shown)


tests/lib/api-request-test.js
```diff
@@ -521,6 +522,64 @@ describe('APIRequest', function() {
...
+			apiRequest.execute(function callback(err, response) {
+				assert.equal(apiRequest.numRetries, 1);
+				assert.equal(err, null);
+				assert.deepEqual(response, expectedResponse);
+				done();
+			});
+
+			setTimeout(function() {
+				requestStub.yieldsAsync(null, responseInfo2);
+			}, 30);
```
mwwoda, line 555: Is it possible to set this mock in a different way without `setTimeout`? Something similar to `onCall` in `token-manager-test.js` to say that on the second call we return a different response?

mwwoda (approved):

PR #808, opened 2023-03-14
feat: Implement shield information barrier

congminh1254:
Closes: SDK-2973
- Implement shield information barrier.
- Updates the Sign Request with newer OpenAPI spec.


codegen/generate-manager-classes.tsx
```diff
@@ -0,0 +1,86 @@
...
+	ImportClause,
+	ImportDeclaration,
+	StringLiteral,
+} from './tsx';
+import { writeNodesToFile } from './write-nodes-to-file';
+tsx;
+
+const SCHEMAS_RELATIVE_PATH = '../src/schemas';
+
+export async function generateManagerClasses(
```
mhagmajer, line 19: It's not clear why changing `generateManagerClass` into `generateManagerClasses` was necessary. In general, the original singular version of the function is arguably more generic/easier to test etc. This is just the previous prototype for codegen that we're not actively maintaining, so no strong opinions here though.

src/schemas/sign-request.generated.ts
```diff
@@ -13,7 +13,12 @@ export interface SignRequest extends schemas.SignRequestCreateRequest {
 	/**
 	 * Array of signers for the sign request
 	 */
-	signers: schemas.SignRequestSigner[];
+	signers?: schemas.SignRequestSigner[];
+	/**
+	 * Force a specific color for the signature (blue, black, or red).
+	 * Example: blue
+	 */
+	signature_color?: string;
```
mhagmajer, line 21: I suppose that the changes to the Sign API that were made since the release aren't reflected in other SDKs either...

mhagmajer (approved):
LGTM and nice tests!

PR #827, opened 2023-05-18
feat: Remove deprecated classes and methods

congminh1254:
In this pull request:
- Remove deprecated classes, methods.
- Update migration guide.


arjankowski (commented):
@congminh1254  is there any reason why you skipped `batch` and `batchExec` from `box-client.ts` in your PR?

There is also a `staleBufferMS` property in config.ts file which is also deprecated. Can you try to remove this as well?

src/box-client.ts
```diff
@@ -634,67 +628,6 @@ class BoxClient {
...
-		if (!this._batch) {
-			return Promise.reject(
-				new Error('Must start a batch before executing')
-			).asCallback(callback);
-		}
-
-		var params = {
-			body: {
-				requests: this._batch.map((batchReq: any /* FIXME */) =>
-					formatRequestForBatch(batchReq.params)
```
arjankowski, line 670: Can you also remove all that is currently not used since you deleted those `batch` method, like:
`formatRequestForBatch` or `_batch`, and maybe more?

arjankowski (approved):
👍

PR #831, opened 2023-05-29
feat: Added support for integration mappings API

arjankowski:


docs/integration-mappings.md
```diff
@@ -0,0 +1,125 @@
+# Integration mappings
```
mgrytsai, line 1: Currently Integration mapping supports only Slack. Suggest to keep the description of this file generic. 

You can get some inspiration from our [dev docs](https://developer.box.com/changelog/#2023-05-24-integration-mappings-API):
> With [integration mappings](https://developer.box.com/reference/resources/integration-mappings/), you can manage where content from partner apps is stored in Box.

This feature currently only supports [Box as Content layer for Slack](https://support.box.com/hc/en-us/articles/4415585987859-Box-as-the-Content-Layer-for-Slack).

mhagmajer (approved):
Looks good!

PR #841, opened 2023-07-14
chore: Update packages to make FIPS compliance

congminh1254:
To support FIPS compliance, Box Node SDK must be able to run on OpenSSL 3.0. This PR bumps some packages that do not support OpenSSL 3.0 and also fix some failing tests.

The new private key used in the test is generated by Box DevPortal but is already deactivated and contains no sensitive information.


mhagmajer (approved):
👍

PR #852, opened 2023-09-18
fix: Overrides tough-cookie dependency

### LINKED ISSUE #851: CVE-2023-26136 because of request@2.88.2/tough-cookie@2.5.0
### Is your feature request related to a problem? Please describe.

Latest box sdk introduces a **critical** security vulnerability ([CVE-2023-26136](https://www.mend.io/vulnerability-database/CVE-2023-26136)) because it still uses requests, which is deprecated for a long time now.

### Describe the solution you'd like

I've seen [this issue](https://github.com/box/box-node-sdk/issues/716), which has simply been closed although nothing was done. It states that a new version of the SDK in Typescript should be released in July/August, we are now in September and it is still not out. 3 months for a critical security vulnerability is an eternity.

I know [this issue](https://github.com/box/box-node-sdk/issues/665) states that it's available as `box-typescript-sdk-gen`, but it's still a beta and:
1. We are really reluctant to use a beta version in a production application
2. This is still a major change, critical security vulnerability should still be patched on existing version IMO

### Describe alternatives you've considered
Forcing the version of touch-cookie with an [override](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) clause, but [version 4 of tough-cookie as breaking change](https://github.com/salesforce/tough-cookie/releases/tag/v4.0.0)s, so I'm scared it would break requests. To be honest with requests being deprecated, it should just be replaced anyway.

There is also [this issue](https://github.com/box/box-node-sdk/issues/537) where someone proposed a replacement for request.

Finally, this issue mention

congminh1254:
- Overrides `tough-cookie` to version 4.1.3.
- Closes #851


mhagmajer (approved):
🚀

PR #863, opened 2023-11-02
feat: Add request option to skip `PassThrough`

rlyonbox:
The `PassThrough` is causing us a few issues, and I can't figure out why it was added - this PR adds an option which avoids the usage of it optionally while leaving the default behavior as PassThrough.


congminh1254 (approved):
LGTM 👍

PR #865, opened 2023-11-03
fix: Update `proxy-agent` usage, drop support Node 12

congminh1254:
Closes: box/boxcli#505

While bumping the `proxy-agent` from version 5 to 6, the import syntax changed from:
```
var ProxyAgent = require('proxy-agent');
```
to
```
import { ProxyAgent } from 'proxy-agent';
```

And proxy-agent is not supporting Node 12 anymore.


mhagmajer (dismissed):
Please also bump the engine version in package.json

mhagmajer (approved):

PR #887, opened 2024-08-23
feat: Support AI agent

congminh1254:


codegen/codegen.tsx
```diff
@@ -188,6 +188,10 @@ import { generateManagerClasses } from './generate-manager-classes';
 						{
 							name: 'textGen',
 							operationId: 'post_ai_text_gen',
+						},
+						{
+							name: 'getDefaultAiAgent',
```
arjankowski, line 193: In generated sdks we named that method `getAiAgentDefaultConfig`. What do you think about keeping same name here?

arjankowski (approved):
🚀

arjankowski (approved):

PR #897, opened 2025-06-12
chore!: Drop support Node 14, 16 and adding support Node 22, 24

congminh1254:


package.json
```diff
@@ -18,7 +18,7 @@
     "sdk"
   ],
   "engines": {
-    "node": ">= 14.0.0"
+    "node": ">= 18.0.0"
```
mwwoda, line 21: I believe we consider this as a breaking change. Please indicate it by including exclamation mark in commit title. 

Also, do you think we should keep it in a separate branch for now?

congminh1254: I added the exclamation mark to the PR title, and yes, we should keep it in separate branch until the major release :D

PR #1042, opened 2025-09-29
feat: Support import syntax in legacy SDK

congminh1254:


Pask423 (approved):
reviewing 117 files is hard but seem ok

